PR #38, opened 2016-04-25
Feature/routing

runarbe:
@wohnout: To the best of my abilities, I have now tried to build a new module using HSLayers. It went comparatively well.

- Added new example that utilizes the SDI4Apps OpenAPI routing service
- Questions to be resolved prior to merge
    - This component uses s4a.js
    - I have simply included s4a.js in the index.html - not via require.js as I am uncertain about the chain
    - The component is hard-coded to work against portal.sdi4apps.eu but could in principle work against any SDI4Apps platform instance
    - The single click event interferes with feature info

On a generic level, how do you isolate application specific modules from the core source code, i.e. the Lodexplorer and the OTN stuff would perhaps mainly be interesting to specific target groups - as would, perhaps, addressing.


raitisbe:
> On a generic level, how do you isolate application specific modules

We write the functions inside app.js or load them there from external js files outside of "components" directory (usualli in "/examples".  Jan is doing it quite extensively in webgl_viz2 example (/home/raitis/Projects/hslayers-ng/examples/webgl_viz2/app.js).

wohnout:
Raitis this will be more like new module for future reuse.

raitisbe:
The code looks great imho. Merging.

PR #365, opened 2019-11-04
Sublayers service

FilipLeitner:


jmacura:
@FilipLeitner Are the one-time bindings (::) necessary here? They are not supported in the new Angular ([see here](https://stackoverflow.com/questions/34375624/angular-2-one-time-binding)), hence they will require more refactoring at some point in the future (when (if ever) we will want to update to Angular >2.x)..

FilipLeitner:
Uhm I see....well, I would say that watchers that are unbinded with :: are not necessary, so as long as we are on Angular 1.x its good to use them where possible. Anyways, the refactoring of this specific thing shouldnt be that problematic as we would have to probably just delete it.

PR #395, opened 2019-11-27
Minisidebar

FilipLeitner:


components/layout/layout.service.js
```diff
@@ -374,7 +380,12 @@ export default ['config', '$rootScope',
         angular.forEach(config.panelsEnabled, (value, key) => {
             me.panelEnabled(key, value)
         })
-
+        me.populateEnabledPanels = function(){
+            if(me.panelsEnabled == 0 && document.getElementsByClassName("sidebar-item").length>0){
```
raitisbe: Do you think theres a way to do this without using elements classes and querying the DOM? Maybe config.panelsEnabled.filter(p => p == true).length or something like that? Using classes is generally bad, because we might have conflict between hslayers classes and container application element classes. When theres realy a need, we might prefix the class with "hs-" to be a bit more safe.

PR #467, opened 2020-01-16
Issue 354/shp2layman

raitisbe:


jmacura:
I must say I don't really like mixing different ES syntax in a single file – vars vs. lets, keyworded functions vs. arrow functions. If you don't mind @raitisbe , I might add some commits to unify the codebase a bit.

But it's not blocking the PR as a whole :-)

raitisbe:
Thanks for pointing out about sld. It realy should be optional

jmacura:
Belatedly, I approve the changes :-) But I propose to let our tester take a look at it now.

PR #471, opened 2020-01-20
Hide search input when search tool in sidebar

FilipLeitner:
Fix: toolbar search input should be hidden when search is enabled in sidebar


FilipLeitner:
will add that with more changes later

raitisbe:
I didnt merge the PR yet. Should we restore?

FilipLeitner:
Well not necessarily, I'll add that with bigger commit later... I tought i wont bother you with just one line so I closed the PR

PR #508, opened 2020-02-12
HSL sizes in container sizemode

FilipLeitner:
- Fix size calculation of hsl map and panelspace when app included in CMS like Wordpress.
- added config parameter for sidebarExpanded state

Further style changes to the panelspace content may be necessary


FilipLeitner:
[Enhance HSL usage in CMS like Wordpress.](https://trello.com/c/iw7ulWCG/15-enhance-hsl-usage-in-cms-like-wordpress)

PR #515, opened 2020-02-20
Geolocation

FilipLeitner:


FilipLeitner:
[Prefix css classes and remove element IDs in geolocation module (refactored)](https://trello.com/c/0eNJ5Q8z/13-prefix-css-classes-and-remove-element-ids-in-geolocation-module-refactored)

PR #528, opened 2020-02-28
Issue 468 changes

FilipLeitner:
Ability to style sidebar from config:
theme: {
sidebar:{
background:'rgb(40, 74, 109)',
itemColor: 'white',
activeItemColor: 'red'
},
-Ability to make all panels as nonQueryable
- Ability to turn off queries on wms layer... could be also added to Add layer component later
-Add possibility to use custom HTML template for displaying query results. To do so set new layer parameter customInfoTemplate: require(your.html) and add customInfo component to app.js injections (to be added eather to examples or incoming template folder suggested by @fzadrazil )
-Some more layout changes/fixes


FilipLeitner:
[allow custom HTML templates for getFeatureInfo response](https://trello.com/c/EHUstLmp/35-allow-custom-html-templates-for-getfeatureinfo-response)

PR #559, opened 2020-03-18
Hot fixes

FilipLeitner:


FilipLeitner:
Uhm, thanks for pointing that out. Ill think of other possibilities.

PR #640, opened 2020-05-07
Don't use "active" class on layermanager subheadings

raitisbe:


raitisbe:
image
Dont you think it looks more easy on the eye this way?

PR #645, opened 2020-05-11
Update ESLint settings

jmacura:
Delete useless .gitkeep
Change angular/typecheck-* rules
Return .vscode/ under VCS
Ignore dist/ in ESLint
Add eslint-plugin-jsdoc

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,7 @@
     "browser": true,
     "es6": true
   },
-  "extends": ["openlayers", "angular"],
+  "extends": ["openlayers", "angular", "plugin:jsdoc/recommended"],
   "globals": {
     "Atomics": "readonly",
     "SharedArrayBuffer": "readonly",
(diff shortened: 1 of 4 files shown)


jmacura:
[Lint modules](https://trello.com/c/6Bg4rPbU/52-lint-modules)

raitisbe:
I think we should not put .vscode under source control because its user speciffic. Can't we have it as a template or put the settings in another file?

jmacura:
See https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control

jmacura:
@raitisbe Having the project-settings shared between collaborators prevents odd misbehaviour.
I don't think we should be using it though. I have no .vscode/ in the HSLayers folder for example. Everything that it can be useful for (line-lenght, line-endings, ...) is already set by eslint rules and now by prettier rules specifically.

raitisbe:
I had exactly the same thought a year ago, but added .vscode to ignore
list because i was using some settings which realy were messing with looks
on Simons computer. Line length guide in a form of vertical line for
example :)

jmacura:
Superceded by #648

PR #663, opened 2020-05-24
Changes for P4A website

jmacura:
These changes are required to present an HSLayers-based map of members in the plan4all website.
It primarily enhances settings and functionality of pop-ups.


jmacura:
@raitisbe I would especially like your review of 2b87abc and b062697, as it is the first unit test I have ever written (and tested) and the changes in utils.service.js might be too ovecomplicated, but I haven't found out more straight-forward approach. Thanks!

PR #681, opened 2020-05-31
Draw GUI + add default attributes based on geom type

FilipLeitner:


FilipLeitner:
[draw GUI](https://trello.com/c/iS7TBKEl/71-draw-gui)

FilipLeitner:
> Features attributes for Area, Length I think should not be stored in feature directly and not in a hardcoded way. Because these attributes get synchronized to server, might be named differently and probably other reasons.
> Dailis already did similar functionality for features center calculation in query module dinamically and just for visualization. See in query-vector.service
> `stats: [{name: 'center', value: toLonLat(getCentroid(feature))}], `

Sounds reasonable.. we can then have the attributes for all queried vectors not jsut the drawn ones. Thanks for pointing it out . Ill change that

FilipLeitner:
[draw GUI](https://trello.com/c/iS7TBKEl/71-draw-gui)

PR #684, opened 2020-05-31
Composition layer visibility

raitisbe:


jmacura:
Seems fine to me
image

PR #686, opened 2020-06-01
Linting cesium

raitisbe:


components/hscesium/cesium-time.service.js
```diff
@@ -1,23 +1,35 @@
 import moment from 'moment';
-const me = {};
 
-export default {
+export class HsCesiumTimeService {
```
jmacura, line 3: with
`export default class`
you can omit the curly braces later on import.
Not sure if there is a reason for named exports.

raitisbe: I took it from some example/tutorial. Maybe the namied export is because then we can see in the service file itself how the service is named and don't need to look it up in XX.module.js in case we forget.

raitisbe: https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

raitisbe: I hope name exports will help in some way when we do the big refacoring and rewriting to angular2

jmacura: Thanks for the link, now it makes perfect sense to me :-)

PR #687, opened 2020-06-01
Update README.md

jmacura:
fix broken links
update text


jmacura:
@raitisbe Actually, this was intentionally just a draft as I wanted to add some more commits later, without the need to create a new branch again.
But that's not really important.

PR #698, opened 2020-06-02
Issue 697/delete features in popup

raitisbe:


raitisbe:
indeed

components/query/feature-popup.component.js
```diff
@@ -34,19 +38,38 @@ export default {
...
+          HsQueryVectorService.removeFeature(feature);
+          HsQueryBaseService.featuresUnderMouse = [];
+        }
       },
-      clearLayer(layer) {
-        layer.getSource().clear();
+
+      async clearLayer(layer) {
+        const confirmed = await HsConfirmDialogService.show(
+          gettext('Really delete all the features in layer?'),
```
jmacura, line 63: I would change this to

`'Really delete all features from this layer?'`

or with template literals it could be

`'Really delete all features from layer "${layer.get('title'}"?'`

raitisbe: Templite literals create a bit of problems with gettext which I dont know how to solve exactly right now.  
`gettext('Really delete all features from layer "{0}"?').replace('{0}', layer.get('title'), ` can work

ChcJohnie: Hi, I had similar problem, possible solution (complete gettextCatalog required) is (adapt to your usecase :) ):
gettextCatalog.getString(
            'Experiment {{name}} model is being computed',
            { name: experiment.name },
          ),

PR #722, opened 2020-06-06
Pre-release 1.24

raitisbe:
# New things:

- Added a calendar datetime picker inside sensor modules chart dialogue.
- New services for controlling Cesium are created for the functionality which was previously not exposed: HsCesiumCameraService, HsCesiumLayersService, HsCesiumTimeService
- Delete feature and clear layer buttons in hover popup window for clustered and unclustered vector layers. Only available if layer is editable - it has  `editor: {editable: true}` in config parameters.
- Universal modal dialog service for yes/no answers accessible using `await HsConfirmDialogService.show(..)`

# Fixed things:

- Generating of thumbnail image fixed when Openlayers uses many canvas elements and with different dimensions.
- visible_layers parameter was overriding layerVisibility and setting it to hidden when layers were loaded from composition.
- Added missing icons to font file and css icon-textfield, icon-edit, icon-time, icon-chevron-down, icon-calendarthree, icon-menu, .icon-navigation, icon-gpson.
- Fixed Cesium crashing when we tried to set its initial viewport to OL maps viewport, but OL map was hidden with dimensions [0, 0] at that time.
- Fixed error ir removePortIfProxified function if proxyPrefix was not set thus port could not be determined.
- Fixed loading of single Tile layers from compositions.
- Fixed checking of duplicate layers, which was previously returning that no duplicates exist.
- Fix duplicity checking of XYZ layers when multiple urls are provided in the embedded layer (Use first one now from the embedded source because composition also only supports one url)
- Hide micka theme dropdown in datasources panel because its empty anyways.

# Changed things:

- Sensor types and groups are now displayed in black.
- Static image layers are now proxied.
- repopulateLayer in HsMapService is now gone and merged into addLayer function.
- Add new layer dialog UI which is called from draw panel and toolbar reworked.
- Drawing panel (not the one in toolbar, but in sidebar) UI improved
- HsDrawServices initiateAddDrawLayer function renamed to saveDrawingLayer and uses already precreated layer with title 'tmpDrawLayer' if functions addNewLayer = false. 
- Sensor lines in charts legend are now sorted by sensor_id  because sorting by name is unreliable when the sensors are named "temperature in 120cm", "temperature in 80 cm" (first is considered smaller).
- HsDrawService.activateDrawing function now uses one options object with attributes instead of many function parameters.
- OSM layers are now always base layers even if not specified as base in composition.
- Feature deletion and layer clearing functions now moved to HsQueryVectorService
- Upgrading to Angularjs 1.8.0
- Updated Czech translations


jmacura:
Thanks for the extensive changelog, @raitisbe !

I am waiting with the approval for the commits from #723 if they are about to come here.

jmacura:
[Release 1.24](https://trello.com/c/TXtHot3S/70-release-124)

raitisbe:
Is there something to add to the release notes from the latest commits or the basically improve on the already described things?

FilipLeitner:
Maybe just
- updated czech translation

PR #723, opened 2020-06-07
Small fixes

FilipLeitner:
- fix transation runblock not working
- update cz translation
- select from listt of coordinates on query instead of showing them all
- fix basemap gallery toggle
- fix adding of default attributes to draw features


FilipLeitner:
[Release 1.24](https://trello.com/c/TXtHot3S/70-release-124)

jmacura:
image

Steps to reproduce:
* open full example and start drawing
* add some features to Bookmarks layer
* add new drawing layer "Drawing layer"
* add some features to Drawing layer
* switch drawing layer back to Bookmarks
* add some features in around existing cluster

FilipLeitner:
> image
> 
> Steps to reproduce:
> 
> * open full example and start drawing
> * add some features to Bookmarks layer
> * add new drawing layer "Drawing layer"
> * add some features to Drawing layer
> * switch drawing layer back to Bookmarks
> * add some features in around existing cluster

Finally found it...just had to go one ste deeper with layer source for clusters. Still clusters doesnt keep selected style but unfortunately I dont have time to fix that it atm.

jmacura:
> Finally found it...just had to go one ste deeper with layer source for clusters. Still clusters doesnt keep selected style but unfortunately I dont have time to fix that it atm.

Very well!
I have tested it and works fine now.

I still encounter invisible drawings in a new drawing layer sometimes, but there is no error in console and I can't find an exact way to reproduce it. So I will look at it once again in the evening and if I will not find any pattern I will approve the PR.

jmacura:
So I think I have caught some silent bugs by 
2012e42 , namely in https://github.com/hslayers/hslayers-ng/pull/723/commits/2012e42e0eb06b97420b6b2d03fd6f0e52f51727#diff-7df3d178ebf66cfbc65b517ed8157384L68 (don't really know what is the result of an assignment inside ternary operator, but it does not seem as a correct way).

There is still a particular issue with circles, which do not display in the map if the layer is clustered.

jmacura:
> There is still a particular issue with circles, which do not display in the map if the layer is clustered - and as long as it is clustered.

Addressed and hopefully fixed in a5b6720 .

Sadly, I have discovered yet another bug meanwhile. It looks like sometimes, a feature is drawn into two layers at once. But don't have time to investigate further now... 😢

PR #726, opened 2020-06-08
Make attribution collapsible by default

FilipLeitner:
Make attribution collapsible by default.

Tried to just modify  existing attribution eg.
  defaultDesktopControls.forEach(control => {
    if (control instanceof Attribution){
      control.setCollapsible(true);
    }
  })
....without success.


FilipLeitner:
Somehow I forgot to change the commit. This was just my first idea beacause I couldnt modify already existing control. Then realized I dont have to do that, just create it with wanted params.

PR #732, opened 2020-06-09
Make Hslayers a hybrid library on top of Angularjs and Angular9

raitisbe:


raitisbe:
[Upgrade to angular 9](https://trello.com/c/3URQG641/76-upgrade-to-angular-9)

raitisbe:
@fzadrazil Yes its possible. From angular upgrade guide https://angular.io/guide/upgrade : "One of the keys to a successful upgrade is to do it incrementally, by running the two frameworks side by side in the same application, and porting AngularJS components to Angular one by one. This makes it possible to upgrade even large and complex applications without disrupting other business, because the work can be done collaboratively and spread over a period of time. The upgrade module in Angular has been designed to make incremental upgrading seamless."

There will be issues for sure, but I like what I see thus far. I was exploring migrating to Angular since we agreed on making parallel branches/tags for the releases and that seems like a good opportunity to test how that works. About semantic versioning Angular9 support seems like a good candidate for 2.x.x version, I just don't know how to go from there when we have some breaking changes in the 1.x.x version, since 2.x would be already taken. I would suggest that we keep 1.24+.x just plain angularjs and do the new developments 
+ in parallel for critical issues
+ in "develop" branch on hybrid with both frameworks.

About estimate, its hard for me since I haven't written any angular 2+ apps. I would guess 3 months, but it's a wild guess. From the app developers perspective there shouldn't be any changes when he uses the hslayersNgConfig function, just the one html <hs> tag. It's a bit more complicated for the app.js use case, but those are basically just us. You can see the changes I made in examples/angular9 branch https://github.com/hslayers/examples/commits/angular9 which involves new tooling for typescript support, new variables for module names instead of 'hs.print' etc.

fzadrazil:
Thanks @raitisbe for the elaboration on the topic. I will go through the changes for sure and I would like to ask @jmacura and @FilipLeitner to do the same. We can talk about it more tomorrow. I was thinking about creating this release-1.24 branch to keep it safe from angular9 and keep the development as you said. But if it takes a few months, I think 1.24.x will not be enough. We need to redesign the Datasources panel (and maybe other components), so I would count with 1.25 at least.

jmacura:
I propose to move this discussion to the Trello ticket as we are going very off-topic to this particular PR.

raitisbe:
Yes

PR #738, opened 2020-06-12
Make attribution collapsible by default

jmacura:
I just realised that we will have to make double PRs when applying changes to AngularJS components – both to release-1.24 branch and to develop.
So I am adding this PR to merge also to develop, otherwise we will lost this commit on the main road.


raitisbe:
Its not possible to do it because map service will be typescript file now.

raitisbe:
Ok for now its still js, but a class

PR #743, opened 2020-06-15
Issue 742/legend ng9

raitisbe:


components/print/partials/printdialog.html
```diff
@@ -1,12 +1,12 @@
 <div class="card panel-default hs-main-panel">
-    <hs-panel-header name="print" title="Print"></hs-panel-header>
+    <hs-panel-header name="print" i18n-title="Print"></hs-panel-header>
```
FilipLeitner, line 2: Might investigate a bit about the internationalization..later, seems like ngx-translate might be a way to go if we want to keep using .po files. https://phrase.com/blog/posts/best-libraries-for-angular-i18n/
Also not sure if its possible to support translations properly in hybrid app. Or does this built i18n way just works right away ?. Dont have an experience with these ng9 things ^^

jmacura:
There are failing unit tests in the services which have been converted to classes.

raitisbe:
> There are failing unit tests in the services which have been converted to classes.

Yeah, not sure if classes or because I changed the bootstraping process. I wIll investigate

components/print/partials/printdialog.html
raitisbe: I was hoping i would work. I would expect that maybe just some module import would be necessary as is with Common module of @angular to tget ngIf and ngModel working. https://angular.io/guide/i18n

raitisbe:
> There are failing unit tests in the services which have been converted to classes.

Tests are now working. Just took 2 full days ;)

PR #774, opened 2020-06-26
Patch release 2.0.1

raitisbe:


raitisbe:
> What issue does it patch? The legend component functionality?

Yes, it was broken for static image layers. And also i forgot to add exjs to the dependecies and because of that i had to include it in the container application, which is not right.

PR #815, opened 2020-07-14
Upgrade measure module to Ng9

jmacura:


jmacura:
[Rewrite measure to ng9](https://trello.com/c/NFoktbfb/86-rewrite-measure-to-ng9)

PR #834, opened 2020-07-23
Draw style

FilipLeitner:
Upgrade Draw and Styler module to ng9


raitisbe:
Looks great

jmacura:
A lot of tricky changes. Good job, @FilipLeitner !

Is it only me or are you also receiving this?
image

raitisbe:
> A lot of tricky changes. Good job, @FilipLeitner !
> 
> Is it only me or are you also receiving this?
> image

Didn't have this, but other error related to context

jmacura:
> Didn't have this, but other error related to context

No? That is strange, because I have found the source of this error :D

FilipLeitner:
> A lot of tricky changes. Good job, @FilipLeitner !
> 
> Is it only me or are you also receiving this?
> image

Havent come across this one either.  If you could describe it more I can also inculde that in unit test

jmacura:
Now I observe an issue with adding new drawing layer. When I try to add it, it does not appear in the Layer Manager and every newly drawn feature disappears when it looses focus.
But no error rises in the console, it's silent.

jmacura:
Steps to reproduce:
* run the full example
* open drawing **toolbar**
* add new drawing layer
* draw first feature
* draw second feature

What happens:
* first feature disappears

What should happen:
* should not disappear

Issue #842, opened 2020-07-27
Rewrite hscesium module to ng9


jmacura:
follow-up to #682

PR #862, opened 2020-07-29
Issue 842/hscesium ng9

raitisbe:
Issue #842


raitisbe:
[Rewrite hscesium module](https://trello.com/c/1oKXCNdQ/90-rewrite-hscesium-module)

components/hscesium/ol-cesium-object-map-item.class.ts
```diff
@@ -0,0 +1,5 @@
+import {DataSource, ImageryLayer} from 'cesium';
+export class OlCesiumObjectMapItem {
```
jmacura, line 2: @raitisbe When you neither intend to create new objects of type `OlCesiumObjectMapItem` with the 'new' keyword nor implement its methods, then I think the interface is a cleaner option (it does not compile into JS and is only used for static type-checking).
The same for the `ParamCacheMapItem` class.
Just a note ;-)

PR #891, opened 2020-08-04
Issue 881/hs layer selector service

FilipLeitner:
+ plus some minor bug fixes

### DIFF
--- a/components/layermanager/layer-editor.service.ts
+++ b/components/layermanager/layer-editor.service.ts
@@ -1,6 +1,7 @@
 import {HsEventBusService} from '../core/event-bus.service';
 import {HsLayerDescriptor} from './layer-descriptor.interface';
 import {HsLayerEditorVectorLayerService} from './layer-editor-vector-layer.service';
+import {HsLayerSelectorService} from './layer-selector.service';
 import {HsLayerUtilsService} from '../utils/layer-utils.service';
 import {HsLayoutService} from '../layout/layout.service';
 import {HsLegendDescriptor} from '../legend/legend-descriptor.interface';
@@ -17,7 +18,6 @@ import {get as getProj, transform, transformExtent} from 'ol/proj';
 })
 export class HsLayerEditorService {
   legendDescriptor: HsLegendDescriptor;
-  currentLayer: any;
 
   constructor(
     private HsMapService: HsMapService,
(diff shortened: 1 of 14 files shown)


components/layermanager/layermanager.component.ts
```diff
@@ -93,10 +93,7 @@ export class HsLayerManagerComponent implements OnInit {
     this.HsMapService.loaded().then((map) => this.init(map));
 
     this.HsEventBusService.layerRemovals.subscribe((layer) => {
-      if (
-        typeof this.HsLayerManagerService.currentLayer == 'object' &&
-        this.HsLayerManagerService.currentLayer.layer == layer
-      ) {
+      if (this.HsLayerManagerService?.currentLayer?.layer == layer) {
```
jmacura, line 96: 🚀

components/layermanager/layer-editor.service.ts
```diff
@@ -26,14 +26,12 @@ export class HsLayerEditorService {
...
-  setLayer(currentLayer: HsLayerDescriptor): void {
-    this.currentLayer = currentLayer;
-    this.legendDescriptor = this.HsLegendService.getLayerLegendDescriptor(
-      currentLayer.layer
-    );
+    private HsLegendService: HsLegendService,
+    private HsLayerSelectorService: HsLayerSelectorService
+  ) {
+    this.HsLayerSelectorService.layerSelected.subscribe((layer) => {
+      this.HsLegendService.getLayerLegendDescriptor(layer.layer);
```
raitisbe, line 33: Assign to legendDescriptor ?

components/layermanager/layer-selector.service.ts
```diff
@@ -0,0 +1,16 @@
...
+import {Subject} from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class HsLayerSelectorService {
+  layerSelected: Subject<any> = new Subject();
+  constructor() {}
+
+  select(layer: Layer) {
```
raitisbe, line 13: Is the type right? Shouldn't it be HsLayerDescriptor instead of ol/Layer, since e are calling layer.layer later on it. Also Subject<HsLayerDescriptor>

components/layermanager/layermanager.component.ts
raitisbe: indeed

components/layermanager/layermanager.service.ts
```diff
@@ -789,6 +791,7 @@ export class HsLayerManagerService {
       layer.layer.checkedSubLayers = {};
       layer.layer.withChildren = {};
     }
+    this.HsLayerSelectorService.layerSelected.next(layer);
```
raitisbe, line 794: this.HsLayerSelectorService.select(layer) would be better I think. From what I understood from the article it's a good practice to hide the subject as a private variable and issue the next() from a public function inside the service (select() in our case) so outside services could not issue the events uncontrolled and create 'event soup'.

FilipLeitner: Indeed, this way I completely left the select method out , not sure why. thanks for pointing it out.

components/layermanager/layer-selector.service.ts
jmacura: This can get very confusing quite quickly. I suspect better variable name (layerDesc or alike) would make the code much clearer 😉

Issue #859, opened 2020-07-28
Nonsense text in WMS layer settings


jmacura:
"Scaleto 1:" should probably be "Min scale 1:123" (Or "Max scale 1:123"?)
![image](https://user-images.githubusercontent.com/5598693/88674373-056a5200-d0ea-11ea-9341-d8872c311be9.png)

Analogically here:
![image](https://user-images.githubusercontent.com/5598693/88674736-76aa0500-d0ea-11ea-806b-015a3eec457e.png)

jmacura:
Not a critical issue, can be fixed in 2.3.0, no need to postpone 2.2.0 any longer.

raitisbe:
Now will be 
![image](https://user-images.githubusercontent.com/395514/88904700-11772080-d25e-11ea-99eb-f1c05637534d.png)

Issue #855, opened 2020-07-27
Point style does not accept decimal line-width


jmacura:
In Styler, when line width is set for Point with non-integer value, the point disappears as a whole.

![image](https://user-images.githubusercontent.com/5598693/88595215-d9f05480-d062-11ea-9841-d4bae0b563fc.png)

![image](https://user-images.githubusercontent.com/5598693/88595227-e1176280-d062-11ea-90a9-f9a674198655.png)

Either allow decimal numbers in line-width param or align the range to integers only.

jmacura:
@fzadrazil Cannot reproduce in 1.24 branch, because we use `parseFloat($scope.linewidth)` instead of `this.linewidth.toFixed(0)` there. Hence there is no need to fix this issue there.

Issue #828, opened 2020-07-20
Uncaught (in Promise): [object Undefined] randomly happening in the console


jmacura:
![image](https://user-images.githubusercontent.com/5598693/87981361-6e900b00-cad5-11ea-8d91-61d670418229.png)

This error sometimes appears in my console (and when it appears, then always twice). It also results in a vector layer which won't get clustered (despite it should be) and an empty Layer Manager.
Can't see where it comes from, as all the files listed as the source of the error in the console are Angular files, not ours.

Gmarquis45:
Same problem here with 2.1.1 version.
Only sometimes, hard to reproduce... :(

raitisbe:
I think its due to loaded() {
    return new Promise((resolve, reject) => {
      if (this.map) {
        resolve(this.map);
        return;
      } else {
        this.$timeout(() => {
          if (this.map) {
            resolve(this.map);
          } else {
            reject();
          }
        }, 1000);
      }
    });
  }
in map.service.js. Map didn't manage to initialize itself in 1000 miliseconds probably. Should write this in a more clever way.

raitisbe:
Easy to catch this error when having debugger; in some service and while it stops the execution, the promise gets rejected.     constructor(private HsMapService: HsMapService, private HsEventBusService: HsEventBusService) {
        debugger;
        this.HsEventBusService.cesiumLoads.subscribe((data) => {
            debugger;
            this.init(data.viewer);
        })
    }

PR #895, opened 2020-08-04
Fix compositions pagination

fzadrazil:
Fixed pager displaying in the compositions panel. We will need to publish release 1.24.2 with this fix. Any ideas what next to include in the release?

### DIFF
--- a/components/compositions/partials/compositions.html
+++ b/components/compositions/partials/compositions.html
@@ -122,7 +122,7 @@
                     </div>
 
 
-                    <div class="card-footer" ng-show='endpoint.compositionsCount>0'>
+                    <div class="card-footer" ng-show='endpoint.compositionsPaging.compositionsCount>0'>
                         <nav>
                             <ul class="pagination d-flex justify-content-center">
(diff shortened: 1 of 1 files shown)


components/compositions/partials/compositions.html
```diff
@@ -122,7 +122,7 @@
                     </div>
 
 
-                    <div class="card-footer" ng-show='endpoint.compositionsCount>0'>
+                    <div class="card-footer" ng-show='endpoint.compositionsPaging.compositionsCount>0'>
```
raitisbe, line 125: Didn't expect it to be this easy :)

raitisbe:
> Fixed pager displaying in the compositions panel. We will need to publish release 1.24.2 with this fix. Any ideas what next to include in the release?

Probably only some bug fixes.
https://github.com/hslayers/hslayers-ng/issues/855
https://github.com/hslayers/hslayers-ng/issues/859
https://github.com/hslayers/hslayers-ng/issues/828
https://github.com/hslayers/hslayers-ng/issues/780
https://github.com/hslayers/hslayers-ng/issues/730
Could be candidates.

PR #896, opened 2020-08-04
Issue 882/save map ng9

raitisbe:


common/endpoints/endpoints.service.ts
```diff
@@ -50,9 +50,10 @@ export class HsCommonEndpointsService {
             user: ds.user,
             liferayProtocol: ds.liferayProtocol,
             originalConfiguredUser: ds.user,
-            getCurrentUserIfNeeded:
-              HsCommonLaymanService.getCurrentUserIfNeeded,
+            getCurrentUserIfNeeded: () =>
+              HsCommonLaymanService.getCurrentUserIfNeeded(tmp),
```
jmacura, line 54: What is this trick? Some nasty 'this' context again?

raitisbe: Pretty much only this time I knew before that it would fail, since I'm passing the function inside the array element/object, but the function itself resides inside the class.

common/widgets/misc.module.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @memberOf hs
+ */
+@NgModule({
+  declarations: [HsMiscRecursiveDd],
+  imports: [CommonModule, BrowserModule],
+  providers: [],
+  entryComponents: [HsMiscRecursiveDd],
+  exports: [HsMiscRecursiveDd],
+})
+export class HsMiscModule {}
```
jmacura, line 16: I believe a better name can be invented for this module+component+service. But I can't found any appropriate now..

components/save-map/sync-error-dialog.html
```diff
@@ -1,14 +1,9 @@
...
-    style="position: fixed; top: 10em; left: calc((100% - 500px) / 2); height: 19em; width: 500px; overflow-y: auto;"
-    ng-class="{'show': $ctrl.modalVisible}" role="alert">
-    <h4 class="alert-heading" translate> 
+<div class="alert alert-danger alert-dismissible fade" style="position: fixed; top: 10em; left: calc((100% - 500),px) / 2; height: 19em; width: 500px; overflow-y: auto;" [ngClass]="{'show': modalVisible}" role="alert">
+    <h4 class="alert-heading" translate="">
         Error when syncing
     </h4>
-    {{$ctrl.exception}}
-    <button type="button" class="close" data-dismiss="alert" aria-label="Close" ng-click="$ctrl.modalVisible = false">
-        <span aria-hidden="true">×</span>
```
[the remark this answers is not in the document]
raitisbe: We do have the span. This is taken from bootstrap example https://getbootstrap.com/docs/4.3/components/toasts/

raitisbe:
> Very well @raitisbe ! I am quite impressed by the recursion in component 😄
> I have left some comments for subtle changes.

Thanks for the review @jmacura . Will try to implement the suggested changes.

components/save-map/sync-error-dialog.html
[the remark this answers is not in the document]
jmacura: But the span is empty in your changes. If I can see it correctly, there is always the `&times;` entity in the bootstrap examples.

common/widgets/misc.module.ts
jmacura: Great, I like this!

components/save-map/sync-error-dialog.html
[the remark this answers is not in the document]
raitisbe: Yes, you are right. Red highlight means that it's removed. Must have been due to the automatic conversion

PR #897, opened 2020-08-05
Issue 864/sidebar to ng9

raitisbe:


raitisbe:
PanelHelpersModule was not a typo :) But you might be right, since now it only deals with Headers it could be called PanelHeaderModule. The plan was to have some more functionality related to side panels.

ajs-app.ts
```diff
@@ -1,40 +1,43 @@
+/* eslint-disable angular/no-controller */
```
jmacura, line 1: Do not disable the eslint rules, unless you really want to mark an exception from the rule. Even though you are not going to fix the errors and warnings right now, just let them shine, so we know that there is some code that should be changed eventually. Thanks.

DailisLangovskis: Got it! I thought that wasn't a big deal, but i will keep that in mind!

PR #913, opened 2020-08-14
New docs

### LINKED ISSUE #744: Fix i18n translations in ng9

jmacura:
The new documentation tool ( @compodoc ) as I have presented at today's telco.

Please check if the old logo has some usage in the apps, but I think it's save to remove it.

Note that this moves Doc generation from Grunt to NPM. Once we will solve also #744 , we can get rid off Grunt for good. 🎊


raitisbe:
Great, thank you!

PR #916, opened 2020-08-17
Issue 841/sourcemap optimization

raitisbe:


raitisbe:
Sourcemaps are still generated for images, but cant figure out how to move to another bundle.
 require(/* webpackChunkName: "img" */'./img/svg/road-sign1.svg') doesnt work obviously
import(/* webpackChunkName: "img" */'./img/svg/road-sign1.svg') kind of does, but only in js files, not ts. 

For ts we would require to use ng-annotate loader with extra ng-annotate-patched plugin if we want to have ngInject for old modules in ts files and we do still.  
Also importing of images in ts is clumsy for libraries such as hslayers because it needs to have a special index.d.ts file in every project which compiles hslayers with content https://stackoverflow.com/questions/51100401/typescript-image-import

Anyways, using this PR the complete bundle size with sourcemaps is reduced from ~25MB to 7Mb.

PR #933, opened 2020-08-19
Issue 865/utils to n9

DailisLangovskis:
Added some small changes based on some errors that occurred.


raitisbe (changes requested):
Not so sure about replaceAll existing in javascript. Maybe just for firefox. See:   console.log('a a a'.replaceAll('a', 'b'));
image
I think we will have to use HsUtilsService.replaceAll(string, what, withWhat)

raitisbe (approved):

PR #936, opened 2020-08-19
Issue 903/custom panels

raitisbe:


jmacura:
Great! I will try to look at this tomorrow.

raitisbe:
@jmacura See https://github.com/hslayers/project-rostenice/commits/upgrade-hslayers-2

jmacura:
image

image

Awesome!

jmacura:
It works just charmingly 😋 
image
image
image

components/layout/panels/panel-component.ts
```diff
@@ -0,0 +1,7 @@
+import {ViewRef} from '@angular/core';
+
+export interface HsPanelComponent {
+  viewRef: ViewRef;
+  data: any;
+  isVisible();
```
jmacura, line 6: add boolean as a required return value type

components/layout/panels/panel-item.class.ts
```diff
@@ -0,0 +1,11 @@
+import {Type} from '@angular/core';
+
+export class HsPanelItem {
+  constructor(public component: Type<any>, public data: any) {}
+}
+
+/*
```
jmacura, line 7: I feel like the following lines can be safely deleted 😉

jmacura (changes requested):
Following our current naming practice, I propose to rename `panel-component.ts` to `panel-component.interface.ts`
and `panel-item.class.ts` to `panel-item.ts`
(reasoning behind that: basically everything is a class now -- all components, services or modules -- but just a few files define interfaces)

**Some general questions:**
Why have you decide to make default panels have `ng-show="panelVisible('name', this)">` defined in the layout.html template, but custom panels must implement the `ng-show` (or `[hidden]`) independently via `isVisible()`?

What is the intended usage of the `data` property in the custom HsPanelComponent?

raitisbe:
> Why have you decide to make default panels have `ng-show="panelVisible('name', this)">` defined in the layout.html template, but custom panels must implement the `ng-show` (or `[hidden]`) independently via `isVisible()`?
> 
Not so much that I did decide, more like I could not find a way how to set the ng-show from outside (from the container component). isVisible in the interface is just so all the components would follow more or less the same principle how to toggle themselves on/off.

raitisbe:
> What is the intended usage of the `data` property in the custom HsPanelComponent?

Pretty much leftover from dialog component, but there are hypothetical use cases, for example having 2 instances of the same panel with various headings or something such.

jmacura (approved):

jmacura:
> 
> 
> > Why have you decide to make default panels have `ng-show="panelVisible('name', this)">` defined in the layout.html template, but custom panels must implement the `ng-show` (or `[hidden]`) independently via `isVisible()`?
> 
> Not so much that I did decide, more like I could not find a way how to set the ng-show from outside (from the container component). isVisible in the interface is just so all the components would follow more or less the same principle how to toggle themselves on/off.

I see. Good choice then, to define the method in the interface.

Issue #461, opened 2020-01-13
3 modes of sidebarPosition in config


raitisbe:
Possible string values: left, right, invisible

This would substitute the current layoutService.sidebarRight layoutService.sidebarLeft properties.

bvj:
And maybe expanded/collapsed state?

raitisbe:
I think that should be another setting

PR #940, opened 2020-08-21
Refactor layout - replace js with css where possible

### LINKED ISSUE #804: Map crashes when resizing window
image

FilipLeitner:
Notes
- hs-app element needs display and height propery to be set
- fixes #804 #461 

 TBD
- delete unused rules in app.css

### DIFF
--- a/components/layout/layout.directive.js
+++ b/components/layout/layout.directive.js
@@ -58,6 +58,9 @@ export default function (HsCore, $timeout, HsConfig, $compile, HsLayoutService) 
           HsLayoutService.layoutElement.style
             .setProperty('--sidebar-active-color', HsConfig.theme.sidebar.activeItemColor || null);
         }
+        if (HsConfig.sidebarRight == false){
+          //sidebarRIghts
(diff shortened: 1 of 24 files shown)


components/layout/layout.directive.js
```diff
@@ -1,64 +1,126 @@
...
+        : require('./partials/layout.html'),
     link: function (scope, element) {
-      HsLayoutService.contentWrapper = element[0].querySelector('.hs-content-wrapper');
+      HsLayoutService.contentWrapper = element[0].querySelector(
+        '.hs-content-wrapper'
+      );
       HsLayoutService.layoutElement = element[0];
 
+      setTimeout(() => {
+        const hsapp = document.getElementById('hs-app');
```
raitisbe, line 29: We should not rely on this id to be set. There can also be multiple hs elements on the page. I propose to just climb back from linked element using parent. It should even be possible in ng9 using components elementRef

raitisbe (approved):

PR #957, opened 2020-08-24
Upgrade datasources to Ng9

jmacura:
This is a first in a series of (at least) 3 PRs, which will hopefully end with a new restyled datasources.

PR #965, opened 2020-08-25
Issue 866/query ng9

raitisbe:


components/core/core.module.ts
```diff
@@ -36,6 +35,7 @@ import {HsToolbarModule} from '../toolbar/toolbar.module';
 import {HsUtilsModule} from './../utils';
 import {HttpClientModule} from '@angular/common/http';
 import {NgModule} from '@angular/core';
+import {TranslateModule} from '@ngx-translate/core';
```
FilipLeitner, line 38: Not sure why but  my webpack throws error with this import. Even in my branch https://github.com/hslayers/hslayers-ng/blob/ngx-translate/components/core/core.module.ts I had to use full path. Does it work for you @jmacura ?

components/query/query-wms.service.ts
```diff
@@ -1,72 +1,103 @@
...
+        Base.setData(group, 'features');
+      }
+    }
+  }
+
+  async request(url, infoFormat, coordinate, layer) {
+    const req_url = this.HsUtilsService.proxify(url, true);
+    const reqHash = this.HsQueryBaseService.currentQuery;
+    try {
+      const response = await this.HttpClient.get(req_url).toPromise();
```
FilipLeitner, line 68: Tried getFeatureInfo request with this WMS: http://geoportal.gov.cz/arcgis/services/CENIA/cenia_corine_2012/MapServer/WmsServer?.
Request was made but the response was not processed as there was no `featureInfoReceived()` call. 

Only managed to get proper response that appears in info panel by replacing toPromise with subscribe. Also XML headers were necessary then
```
    try {
      const headers = new Headers({'Content-Type': 'text/xml'});
      headers.set('Accept', 'text/xml');
      headers.set('Content-Type', 'text/xml');
      await this.HttpClient.get(req_url, {
        headers: new HttpHeaders().set('Content-Type', 'text/xml'),
        responseType: 'text',
      }).subscribe((response) => {
        if (reqHash != this.HsQueryBaseService.currentQuery) {
          return;
        }
        this.featureInfoReceived(response, infoFormat, url, coordinate, layer);
      });
    } 
```

FilipLeitner:
Tried getFeatureInfo request with this WMS: http://geoportal.gov.cz/arcgis/services/CENIA/cenia_corine_2012/MapServer/WmsServer?.
Request was made but the response was not processed as there was no `featureInfoReceived()` call. 

Only managed to fix it 
    try {
      const headers = new Headers({'Content-Type': 'text/xml'});
      headers.set('Accept', 'text/xml');
      headers.set('Content-Type', 'text/xml');
      await this.HttpClient.get(req_url, {
        headers: new HttpHeaders().set('Content-Type', 'text/xml'),
        responseType: 'text',
      }).subscribe((response) => {
        if (reqHash != this.HsQueryBaseService.currentQuery) {
          return;
        }
        this.featureInfoReceived(response, infoFormat, url, coordinate, layer);
      });
    } catch (err) {

components/query/query-base.service.ts
```diff
@@ -0,0 +1,403 @@
...
+          feature.getLayer(map) &&
+          feature.getLayer(map).get('title').length > 0 &&
+          feature.getLayer(map).get('title') !== 'Point clicked'
+        );
+      });
+      this.featureLayersUnderMouse = this.featuresUnderMouse.map((f) =>
+        f.getLayer(this.HsMapService.map)
+      );
+      this.featureLayersUnderMouse = this.HsUtilsService.removeDuplicates(
+        this.featureLayersUnderMouse,
```
FilipLeitner, line 147: Pop up seems to be working.... we just might also filter out features that doesnt have `.get('popUp') `...  (thats where popUp content come from, right?) so we doesnt show empty popup on drawn features or features that come from layer not configured to show pop-ups

components/core/core.module.ts
FilipLeitner: Was talking to Dailis about this one and it seems like error on my side...nevertheless according to docs it is recommended to use  12.x+ version of ngx-translate/core with Angular 9.

components/query/query-base.service.ts
jmacura: Well, `.get('popUp')` might be used, but on a `Layer` level. The attributes to be displayed are described there, not on a feature level.

components/query/query-wms.service.ts
raitisbe: Thanks @FilipLeitner for giving me some pointers. I changed it to {'Content-Type': 'text'} to support wider range of info_formats. Also toPromise() works for me now.

components/core/core.module.ts
jmacura: > 
> 
> Not sure why but my webpack throws error with this import. Even in my branch https://github.com/hslayers/hslayers-ng/blob/ngx-translate/components/core/core.module.ts I had to use full path. Does it work for you @jmacura ?

Nice and clean. No error anywhere.

components/query/partials/feature-popup.html
```diff
@@ -1,47 +1,43 @@
...
-                        <button class="btn btn-danger float-right btn-sm" ng-if="isFeatureRemovable(subFeature)" ng-click="removeFeature(subFeature)"
-                            translate data-toggle="tooltip" title="{{'Remove feature' | translate}}">
+                        <button class="btn btn-danger float-right btn-sm" *ngIf="isFeatureRemovable(subFeature)" (click)="removeFeature(subFeature)" i18n data-toggle="tooltip" i18n-title="Remove feature">
                             <i class="icon-remove-circle"></i>
                         </button>
-                        <div ng-repeat="attribute in subFeature.attributesForHover">
+                        <div *ngFor="let attribute of subFeature.attributesForHover">
                             <div class="d-flex flex-row">
                                 <div class="p-1 flex-grow-1">
-                                    {{attribute.key}}&emsp;
```
jmacura, line 16: I believe the `&emsp;` entity was quite intentional here. Otherwise the key and the value could be rendered as `keyvalue`.

components/query/partials/feature-popup.html
```diff
@@ -1,47 +1,43 @@
...
                                 </div>
-                                <div class="p-1" ng-bind-html="attribute.displayFunction(attribute.value)"></div>
+                                <div class="p-1" [innerHTML]="attribute.displayFunction(attribute.value)"></div>
                             </div>
                         </div>
                         <hr>
                     </div>
                 </div>
-                <div ng-show="!isClustered(feature) && feature.attributesForHover.length > 0" style="font-size: small; line-height: 1; margin-bottom: 2px; border-bottom: 1px solid rgb(199, 199, 199)">
+                <div [hidden]="!(!isClustered(feature) && feature.attributesForHover.length > 0)" style="font-size: small; line-height: 1; margin-bottom: 2px; border-bottom: 1px solid rgb(199, 199, 199)">
```
jmacura, line 23: Double negation is not necessary I believe.
This specific line shall be `isClustered(feature) || !(feature.attributesForHover.length > 0)`. But careful, the second part can not be converted to `feature.attributesForHover.length <= 0` safely, because it would not catch the situation when `feature.attributesForHover.length` is undefined.

components/query/partials/feature-popup.html
```diff
@@ -1,47 +1,43 @@
...
-                        ng-if="isFeatureRemovable(feature)"
-                        ng-click="removeFeature(feature)" translate data-toggle="tooltip" title="{{'Remove feature' | translate}}">
+                    <button class="btn btn-danger float-right btn-sm" *ngIf="isFeatureRemovable(feature)" (click)="removeFeature(feature)" i18n data-toggle="tooltip" i18n-title="Remove feature">
                         <i class="icon-remove-circle"></i>
                     </button>
-                    <div ng-repeat="attribute in feature.attributesForHover">
+                    <div *ngFor="let attribute of feature.attributesForHover">
                         <div class="d-flex flex-row">
                             <div class="p-1 flex-grow-1">
-                                {{attribute.key}}&emsp;
```
jmacura, line 34: `&emsp;` intentional again

components/query/partials/infopanel.html
```diff
@@ -1,22 +1,22 @@
 <hs-panel-header name="info" title="Info"></hs-panel-header>
-<div class="card panel-default hs-main-panel" ng-show="!data.coordinates[0].name.length > 0">
-    <div class="card-body m-2" translate="">Click on map for query results!</div>
+<div class="card panel- hs-main-panel" *ngIf="HsQueryBaseService.data.coordinates.length == 0">
```
jmacura, line 2: great! But `HsQueryBaseService.data.coordinates.length == 0` won't catch `HsQueryBaseService.data.coordinates.length == undefined`. It might be OK, if the undefined cannot happen by other means.

jmacura, line 2: Useless `panel-` again.

jmacura (changes requested):

components/query/feature.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  isFeatureRemovable() {
+    if (this.feature.feature) {
+      return this.HsQueryVectorService.isFeatureRemovable(this.olFeature());
+    } else {
+      return false;
+    }
+  }
+
+  saveNewAttribute(attributeName, attributeValue) {
+    if (this.feature && this.feature.feature) {
```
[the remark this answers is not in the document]
DailisLangovskis: Will be using `this.feature.feature !== undefined` instead

components/query/partials/default-info-panel-body.html
```diff
@@ -1,22 +1,19 @@
-<div ng-show="data.coordinates[0].name.length > 0" class="card panel-default hs-main-panel">
+<div [hidden]="!(HsQueryBaseService.data.coordinates.length > 0)" class="card panel- hs-main-panel">
```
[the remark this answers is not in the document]
DailisLangovskis: classes removed

components/query/partials/feature-popup.html
DailisLangovskis: EN space returned

components/query/partials/feature-popup.html
DailisLangovskis: Em space returned

components/query/partials/infopanel.html
DailisLangovskis: Array HsQueryBaseService.data.coordinates: [] will always be defined, because it is being instantiated  on service initialization.   But if check also for undefined is needed then only this works => HsQueryBaseService.data.coordinates === undefined || HsQueryBaseService.data.coordinates.length == 0

DailisLangovskis: Removed

components/query/partials/feature-popup.html
DailisLangovskis: Replaced to code as suggested, yet because of few popup errors couln't check if everything works as expected.

components/query/feature.component.ts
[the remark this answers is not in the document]
jmacura: I don't recommend that. It would throw an error when `this.feature` equals undefined or null.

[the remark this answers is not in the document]
DailisLangovskis: Hmm... I did try Your suggestion,(`this.feature?.feature`) yet it returned undefined as well

jmacura:
Drawing new features on this branch triggers an error:
image

components/query/feature.component.ts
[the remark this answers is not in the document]
DailisLangovskis: Probably should leave this like it is, cannot think of any better check  right now. <?.> in not working

DailisLangovskis:
> Drawing new features on this branch triggers an error:
> image

fixed!

components/query/partials/feature-popup.html
DailisLangovskis: fixed!

components/query/feature.component.ts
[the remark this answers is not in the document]
raitisbe: Maybe we should also target es2019 as per https://stackoverflow.com/questions/58725711/how-to-get-optional-chaining-working-in-typescript . In case we want to drop babel in future.

DailisLangovskis:
https://github.com/hslayers/hslayers-ng/pull/965/commits/d651b9915d7d953923d61656033f76c48c7a6263  resolves #1011

components/query/feature.component.ts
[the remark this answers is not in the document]
jmacura: The thing is, we already use optional chaining in our TS files (e.g. https://github.com/hslayers/hslayers-ng/blob/develop/components/layermanager/layermanager.component.ts#L97) and it works like a charm. Targeting es5 is a good setting in my opinion. It compiles the TS code into a pretty basic JS, which will secure that hslayers will run (somehow) even in older browsers. Which might be important for mobile browsers for instance.

@DailisLangovskis If it is causing you too much pain (as I can see it does :-) ) just drop my suggestion, which is more or less stylistic and go on. This condition is no game changer. There are more important things to do I believe.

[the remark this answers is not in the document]
raitisbe: @jmacura Right, I thought that we were targeting esnext but it was in the lib option in tsconfig. Didn't check first. In that case I onder hy it didnt work before, because we are having the chaining in ts file. Will have to make a test for this so we don't break the optional chaining in future with some tsconfig/webpack change.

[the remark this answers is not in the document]
DailisLangovskis: @jmacura I am not bothered by this what so ever, just wanted to make sure, it if working fine. Now I only want to understand, do we need the babel or not? I just did this in my spare to for educational purposes :)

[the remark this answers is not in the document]
jmacura: @DailisLangovskis As far as I can tell, we only need Babel for the rest of the JS files, until we rewrite them to TS. TypeScript compiler (or `ts-loader` in our case) will do the job afterwards.

components/query/query-base.service.ts
FilipLeitner: Sure, freatures where feature.getLayer(map).get('popUp') is false

raitisbe: I resolved it here: https://github.com/hslayers/hslayers-ng/pull/965/commits/e5909c3439e7343e0fcd51d58fd3d74313f8b0d3

ajs-upgraded-providers.ts
```diff
@@ -102,32 +102,6 @@ export const HsDimensionServiceProvider = {
...
 
-/**
- * @param i
- */
-export function hsQueryBaseService(i: any) {
-  return i.get('HsQueryBaseService');
-}
-
-export const HsQueryBaseServiceProvider = {
-  provide: HsQueryBaseService,
```
jmacura, line 113: You can safely remove the imports of `HsQueryBaseService` and `HsQueryVectorService` from the top of the file.

common/confirm/confirm-dialog.html
```diff
@@ -1,18 +1,18 @@
-<div ng-class="{'in': $ctrl.modalVisible}" class="modal" tabindex="-1" role="dialog" aria-hidden="true">
+<div class="modal in" tabindex="-1" role="dialog" aria-hidden="true">
```
jmacura, line 1: I guess that "in" class is just a relict without any usage

components/datasource-selector/partials/datasource_selector.html
```diff
@@ -1,4 +1,4 @@
-<div class="card panel-default hs-main-panel">
+<div class="card hs-main-panel">
```
jmacura, line 1: Ah, you want some merge conflicts with #957 ? 😆

components/query/default-info-panel-body.directive.ts
```diff
@@ -0,0 +1,14 @@
+import {Component, OnInit} from '@angular/core';
+import {HsQueryBaseService} from './query-base.service';
+
+@Component({
+  selector: 'hs-query-default-info-panel-body',
+  template: require('./partials/default-info-panel-body.html'),
+})
+export class HsQueryDefaultInfoPanelBodyComponent implements OnInit {
```
jmacura, line 8: What about renaming this file to `default-info-panel-body.component.ts` ?

components/query/partials/feature-popup.html
jmacura: Strange, but I can see it as missing still (replaced by regular space)...

components/query/partials/feature-popup.html
jmacura: Same as above.

common/confirm/confirm-dialog.html
raitisbe: I think its needed so the modal doesn't 'open' in invisible state. Bootstrap things.

components/datasource-selector/partials/datasource_selector.html
raitisbe: Don't think I can do much a bout it now. Could create another commit which puts back the class, but the merge conflict would still happen at some moment when rebasing.

components/query/default-info-panel-body.directive.ts
raitisbe: Can do

raitisbe:
I thinks its fine that &emsp; is missing. We can use flexbox mx-1 or similar classes to shift the value to right. Maybe p-1 is enough already

ajs-upgraded-providers.ts
raitisbe: Done

raitisbe:
> image
> And it gets fired like billion times
> 
> **UPDATE:**
> Steps to reproduce:
> 
> 1. select any feature
> 2. scroll the map

Been having some issues with this for a while. I made a bit more radical rewrite in https://github.com/hslayers/hslayers-ng/pull/965/commits/38374b9c9be98165309b943cdb3dd4ba4064d37b . Now also have to make it perform better, because getLayerForFeature is called on each mouse move and it loops through all features in all layers hich hits performance pretty bad.

components/datasource-selector/partials/datasource_selector.html
jmacura: It's fine. I will resolve the merge conflict when rebasing.

raitisbe:
I can't find where the remaining changes requested are. @jmacura can you help?

jmacura:
@raitisbe Probably the missing `&emsp;` in `feature-popup.html` are still pending...?

raitisbe:
> popup

I dont think we need that &emsp;

jmacura:
> 
> 
> > popup
> 
> I dont think we need that

Ok, we probably don't.
But while checking its layout, I have discovered there is an issue with the popUp config. Namely, the `attributesForHover` array is empty.

jmacura (approved):
Shall be OK now.

raitisbe:
> Shall be OK now.

Thank you

PR #967, opened 2020-08-25
Release 2.3.0

jmacura:
As mentioned at the last Friday's telco, I propose to make a new release in the 2.x series.
We already have quite a lot of changes in the code.
I would also make good use of it in the polirural clustering app.

Only #958, #940 and dependency updates shall be merged before this release. Other PRs can part of the next one.

PR #978, opened 2020-08-26
Fix merge fail where hs.permalink should be hs-share

raitisbe:


raitisbe:
Should me berged before https://github.com/hslayers/hslayers-ng/pull/967

jmacura (approved):
Of course.

PR #1004, opened 2020-09-02
Fix URL generation

### LINKED ISSUE #998: Permalink URL generation is appending into URL instead of updating params
The URL is getting longer and longer while the user interacts with the app. The reason is that the parametres are not replaced by a new value but the whole URL fragment is appended to the end of the URL again.

jmacura:
partially resolves #998 

The URL won't get appended but updated.
? and # marks are treated correctly.
Existing URL gets parsed as expected.
Yet the history of browsing is not acting as expected (moving through browser history does not change the view).

### DIFF
--- a/components/permalink/share-url.service.ts
+++ b/components/permalink/share-url.service.ts
@@ -233,7 +233,7 @@ export class HsShareUrlService {
    * @memberof HsPermalinkUrlService
    * @param {object} key Key name for pushed parameter
    * @param {object} new_value Value for pushed parameter
-   * Push new key-value pair into paramater object and update Url string with new params
+   * @description Push new key-value pair into paramater object and update Url string with new params
    */
   push(key, new_value): void {
     this.params[key] = new_value;
@@ -290,7 +290,6 @@ export class HsShareUrlService {
         this.HsUtilsService.debounce(
           (data) => {
             this.update();
-            //this.browserUrlUpdated.next();
           },
           200,
           false,
@@ -311,7 +310,6 @@ export class HsShareUrlService {
           }
           timer = setTimeout(() => {
             this.update();
-            //this.browserUrlUpdated.next();
           }, 1000);
         });
       });
--- a/components/permalink/share-url.service.ts
+++ b/components/permalink/share-url.service.ts
@@ -97,7 +97,7 @@ export class HsShareUrlService {
     this.HsUtilsService.debounce(
       () => {
         this.Location.replaceState(
-          this.Location.path(),
+          '',
           Object.keys(this.params)
             .map((key) => {
               return {key, value: this.params[key]};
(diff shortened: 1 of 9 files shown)


components/permalink/share.module.ts
```diff
@@ -23,7 +24,9 @@ import {WINDOW_PROVIDERS} from '../utils/window';
   providers: [
     HsShareService,
     HsShareUrlService,
-    {provide: LocationStrategy, useClass: HashLocationStrategy},
+    Location,
+    {provide: LocationStrategy, useClass: PathLocationStrategy},
```
raitisbe, line 28: Thanks. I should have been more careful with copy/paste of strategy.

components/permalink/share-url.service.ts
```diff
@@ -279,18 +280,17 @@ export class HsShareUrlService {
...
+  init(map: Map): void {
     if (this.url_generation) {
       let timer = null;
-      // eslint-disable-next-line angular/on-watch
       this.HsEventBusService.mapExtentChanges.subscribe(
         this.HsUtilsService.debounce(
           (data) => {
             this.update();
-            this.browserUrlUpdated.next();
+            //this.browserUrlUpdated.next();
```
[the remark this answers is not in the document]
raitisbe: Nevermind, I see it now, double issuing of next()

raitisbe (approved):

components/permalink/share-url.service.ts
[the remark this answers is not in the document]
jmacura: Yes, it is broadcasted inside the update(), so it seems redundant. But wasn't sure about it, so I left it inside the comment. I will delete it now when I know it is safe.

jmacura:
Ready to be merged.

components/permalink/share-url.service.ts
```diff
@@ -94,7 +97,7 @@ export class HsShareUrlService {
     this.HsUtilsService.debounce(
       () => {
         this.Location.replaceState(
-          this.Location.path(),
```
raitisbe, line 97: Seems I was too quick to approve this. Now we have a problem that if the page is located in a subdirectory, hslayers rewrites the url by removing the directory, for example https://local.polirural-mb.lv/private/?hs_x=28... becomes https://local.polirural-mb.lv/?hs_x=2865324.91..

raitisbe: Fixed this in https://github.com/hslayers/hslayers-ng/pull/1012/commits/cd17e1ae933dc81dc20f6c9dd7a5b796a5908194

jmacura: Oh, I didn't realised that this can happen. I was too quick in my assumptions too.

PR #1012, opened 2020-09-03
Composition parser class

raitisbe:


jmacura:
Well, maybe it behaves differently now with the correct LocationStrategy, but previously the `Location.path()` was returning the get params **only**.

components/permalink/share-url.service.ts
```diff
@@ -142,6 +141,10 @@ export class HsShareUrlService {
     }
   }
 
+  private pathName(): string {
```
jmacura, line 144: smart and simple 👍

PR #1048, opened 2020-09-15
Bug fixes for groundwater project

raitisbe:


FilipLeitner (approved):
Wouldnt expect that querySelector doesnt work for hs-chartplace

PR #1052, opened 2020-09-16
Layman wfs sync

raitisbe:


raitisbe:
https://drive.google.com/file/d/1NOF66HYtcPZpqecAuuj6UULCDoy_zgEn/view

raitisbe:
Modifying features still doesn't sync

components/save-map/layer-synchronizer.service.ts
```diff
@@ -37,10 +39,23 @@ export class HsLayerSynchronizerService {
...
     this.crs = map.getView().getProjection().getCode();
     this.HsLaymanService.crs = this.crs;
   }
 
+  private async reloadLayersOnAuthChange() {
+    for (const layer of this.syncedLayers) {
+      const layerSource = layer.getSource();
+      layer.set('events-suspended', true);
+      layerSource.clear();
+      await this.pull(layer, layerSource);
```
FilipLeitner, line 54: On logout this would return no features, right? If so we could maybe also remove layer from layermanager, or is it automatic through ol.layer remove event?

raitisbe: Seems a bit too complicated. I would have to add the layer back on login afterwards. Also I'm populating the layer with some default values from anonymous 'browser' user when its not authorized yet.

common/layman/layman-login.component.js
```diff
@@ -3,10 +3,10 @@ export default {
   bindings: {
     url: '=',
   },
-  controller: function ($scope) {
+  controller: function ($scope, HsCommonLaymanService) {
     'ngInject';
     this.modalVisible = true;
-    $scope.$on('datasource-selector.layman_auth', () => {
+    HsCommonLaymanService.authenticated.subscribe((endpoint) => {
```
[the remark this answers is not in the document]
raitisbe: @FilipLeitner thanks. Autorename didn't quite follow through in js files.

FilipLeitner (approved):

PR #1066, opened 2020-09-21
Sanitize styler icons and rework styler

### LINKED ISSUE #1010: Feature styler svg images are not being loaded
icons

raitisbe:
fixes #1010

### DIFF
--- a/components/layermanager/layer-editor-vector-layer.service.ts
+++ b/components/layermanager/layer-editor-vector-layer.service.ts
@@ -8,6 +8,7 @@ import {HsUtilsService} from '../utils/utils.service';
 import {Injectable} from '@angular/core';
 import {Layer} from 'ol/layer';
 import {Point} from 'ol/geom';
+import {createDefaultStyle} from 'ol/style/Style';
 
 @Injectable({
   providedIn: 'root',
@@ -115,39 +116,16 @@ export class HsLayerEditorVectorLayerService {
     resolution: number,
     layer: VectorLayer
   ) {
-    let originalStyle;
     const featureStyle = feature.get('features')
       ? feature.get('features')[0].getStyle()
       : feature.getStyle();
-    if (featureStyle) {
-      originalStyle = featureStyle;
-    } else {
-      if (layer.get('hsOriginalStyle')) {
-        originalStyle = layer.get('hsOriginalStyle');
-      } else {
-        const defaultStyle = new Style({
-          stroke: new Stroke({
-            color: 'rgba(0, 153, 255, 1)',
-            width: 1.25,
-          }),
-          fill: new Fill({
-            color: 'rgba(255,255,255,0.5)',
-          }),
-          image: new Circle({
-            radius: 5,
-            fill: new Fill({
-              color: 'rgba(255,255,255,0.5)',
-            }),
-            stroke: new Stroke({
-              color: 'rgba(0, 153, 255, 1)',
-              width: 1.25,
-            }),
-          }),
-        });
-        originalStyle = defaultStyle;
-        layer.set('hsOriginalStyle', defaultStyle);
-      }
-    }
+
+    const originalStyle = featureStyle
+      ? featureStyle
+      : layer.get('hsOriginalStyle')
+      ? layer.get('hsOriginalStyle')
+      : createDefaultStyle;
+
     let appliedStyle = this.applyStyleIfNeeded(
       originalStyle,
       feature,
--- a/components/layermanager/layer-editor-vector-layer.service.ts
+++ b/components/layermanager/layer-editor-vector-layer.service.ts
@@ -65,10 +65,8 @@ export class HsLayerEditorVectorLayerService {
    * @param {number} distance
    */
   cluster(newValue: boolean, layer: Layer, distance: number): void {
-    if (!layer.get('hsOriginalStyle')) {
-      layer.set('hsOriginalStyle', layer.getStyle());
-    }
     if (newValue == true && !layer.get('declutter')) {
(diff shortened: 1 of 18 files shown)


raitisbe:
[Release 2.4.0](https://trello.com/c/ZPfH6ews/106-release-240)

raitisbe:
After sanitization I tried both with url-loader and file-loader. Still can't explain why it was working in other projects without sanitization before.

components/layermanager/partials/layermanagermd.html
```diff
@@ -99,7 +99,7 @@
                     ng-if="layer.layer.getSource().styleAble">{{'LAYERMANAGER.layermanagerMd.layerStyle' | translate}}</label>
                 <div ng-if="layer.layer.getSource().hasPoly">
                     <div md-color-picker ng-model="layer.style.fillColor" label="Fill"
-                        ng-change="HsLayerEditorStylesService.HsLayerEditorStylesService.saveStyle(layer)"
```
jmacura, line 102: Well spotted!

jmacura (changes requested):
I really like the way you break up large functions into smaller private methods. The code is more readable then 👍 

I have tested the full example under this branch. The icons now do appear as they should.
But there is another issue now: the styler won't restyle features (neither polygons nor lines) until I click the "Icon" button in the Point section. The restyling of points happens immediately, as expected. But the "Circle" option doesn't seem to show anything at all...

DailisLangovskis:
> I really like the way you break up large functions into smaller private methods. The code is more readable then 👍
> 
> I have tested the full example under this branch. The icons now do appear as they should.
> But there is another issue now: the styler won't restyle features (neither polygons nor lines) until I click the "Icon" button in the Point section. The restyling of points happens immediately, as expected. But the "Circle" option doesn't seem to show anything at all...

Fixed!

jmacura:
> 
> > But there is another issue now: the styler won't restyle features (neither polygons nor lines) until I click the "Icon" button in the Point section. The restyling of points happens immediately, as expected. But the "Circle" option doesn't seem to show anything at all...
> 
> Fixed!

Thanks, a follow-up issue: all styling works, until I change "Level" from "Layer" to "Feature" and then back again. Then no styling works at all.

jmacura:
The unit-tests are failing because of missing mock of HsUtilsService in styler.spec.ts. But you probably know that, right? 🙂

jmacura:
> 
> Thanks, a follow-up issue: all styling works, until I change "Level" from "Layer" to "Feature" and then back again. Then no styling works at all.

I have tested it a bit more and it is a bit more clear now: The bug only happens with layers which are not clustered.

raitisbe:
> layers which are not clustered.

Dhould be fixed now

jmacura:
After the last commits: It works fine for "Circle" style now, but still not for "Icon" style.

My commits have added few more defaults.

raitisbe (commented):
I think we can also read circle/icon type from the existing style same way as we do for color. Could be another issue together with reading of alpha, icon size and other properties.

raitisbe (commented):
Well spotted

components/layermanager/layer-editor-vector-layer.service.ts
```diff
@@ -115,39 +116,16 @@ export class HsLayerEditorVectorLayerService {
...
-              color: 'rgba(255,255,255,0.5)',
-            }),
-            stroke: new Stroke({
-              color: 'rgba(0, 153, 255, 1)',
-              width: 1.25,
-            }),
-          }),
-        });
-        originalStyle = defaultStyle;
-        layer.set('hsOriginalStyle', defaultStyle);
```
raitisbe: Can't set hsOriginalStyle as many times as there are features

components/styles/styler.component.ts
```diff
@@ -25,21 +25,8 @@ type StyleJson = {
 })
 export class HsStylerComponent {
   icons: any;
-  imagetypes: Array<any> = [
```
raitisbe, line 25: Originally this approach was used to support translations of "Icon"and "Circle". hrname could change, but underlying value would remain,.

jmacura: Ah, I did not realised that.. Will revert my "improvement" then.

components/styles/styler.service.ts
```diff
@@ -62,4 +63,20 @@ export class HsStylerService {
       }),
     ];
   }
+
+  /**
+   * @function save
+   * @memberof hs.styler.controller
+   * @description Get current style variables value and style current layer accordingly
```
FilipLeitner, line 70: Incorrect description. Would change tjhat ust dont want to interfeere with unexpected commits

jmacura: I have some commits on the go. I gonna squeeze this change between them 😉 Thanks for the tip.

components/styles/styler.component.ts
jmacura: Reverted.

components/styles/styler.service.ts
jmacura: Corrected in cbdaab6

jmacura (approved):
Seems good enough for me now.

Just a last question: is there a reason, why lines and polygons are styled with the same line-color and line-width, while points have their own style?

raitisbe:
If I rememer correctly its because icon in openlayers expects a seperate config of colors and lines.

DailisLangovskis:
> If I rememer correctly its because icon in openlayers expects a seperate config of colors and lines.

This is true, icon and circle uses image style which has its own stroke and fill colors

PR #1080, opened 2020-09-24
Allow cluster styling

FilipLeitner:
Wanted to add to #1066 just forgot to send comment ^^

### DIFF
--- a/components/styles/styler.component.ts
+++ b/components/styles/styler.component.ts
@@ -255,6 +255,7 @@ export class HsStylerComponent {
         );
         this.HsLayerEditorVectorLayerService.styleLayer(layer);
         break;
+      default:
       case 'layer':
         this.setStyleForFeatures(layer, null);
         if (this.isClustered) {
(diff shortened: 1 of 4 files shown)


components/styles/styler.component.ts
```diff
@@ -233,24 +241,33 @@ export class HsStylerComponent {
...
-        if (isClustered) {
+        if (this.isClustered) {
           /* hsOriginalStyle is used only for cluster layers 
           when styling clusters with just one feature in it */
           this.HsStylerService.layer.set('hsOriginalStyle', style);
         } else {
           layer.setStyle(style);
         }
         break;
+      default:
```
raitisbe, line 268: I think having default option 'layer' was fine.

FilipLeitner: Default applies when no other condition applies so when we have 'level: cluster | feature | layer = layer' , layer is still default. But surely I can change the code style

FilipLeitner: 5b52b0c

raitisbe (approved):

PR #1083, opened 2020-09-25
Draw adjustments

FilipLeitner:
Mostly changes requested by Petr Vacha

...There have also been a debate about the drawing finishing using right click. Whether ill be more intuitive to preserve last sketch vertex (add it to the feature before finishing drawing)  or to preserve only the 'stable' ones  (ones that are added by left click).

In the PR the first option is used.  On the other hand tester who tested it on v1.24 asked for the second one.

### DIFF
--- a/components/draw/draw.service.ts
+++ b/components/draw/draw.service.ts
@@ -375,6 +375,28 @@ export class HsDrawService {
         source: this.source,
         type: /** @type {GeometryType} */ this.type,
         style: changeStyle ? changeStyle() : undefined,
+        condition: (e) => {
+          if (e.originalEvent.buttons === 1) { //left click
+            return true;
+          } else if (e.originalEvent.buttons === 2) { //right click
+            if (this.type == 'Point') {
+              return false;
+            } else if (this.type == 'Polygon') {
+              const vertexCount = this.draw.sketchLineCoords_.length;
+              if (vertexCount >= 3) {
+                setTimeout(() => {
+                  this.draw.finishDrawing();
+                }, 250);
+                return true;
+              }
+              return false;
+            }
+            setTimeout(() => {
+              this.draw.finishDrawing();
+            }, 250);
+            return true;
+          }
+        },
       });
 
       this.draw.setActive(drawState);
--- a/components/draw/draw.service.ts
+++ b/components/draw/draw.service.ts
@@ -4,7 +4,9 @@ import Collection from 'ol/Collection';
 import VectorLayer from 'ol/layer/Vector';
 import {Circle, Fill, Stroke, Style} from 'ol/style';
 import {Draw, Modify} from 'ol/interaction';
+import {HsCommonEndpointsService} from '../../common/endpoints/endpoints.service';
 import {HsConfig} from '../../config.service';
+import {HsConfirmDialogComponent} from './../../common/confirm/confirm-dialog.component';
 import {HsDialogContainerService} from '../layout/dialogs/dialog-container.service';
 import {HsDrawLayerMetadataDialogComponent} from './draw-layer-metadata.component';
 import {HsEventBusService} from '../core/event-bus.service';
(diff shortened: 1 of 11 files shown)


raitisbe:
> Mostly changes requested by Petr Vacha
> 
> ...There have also been a debate about the drawing finishing using right click. Whether ill be more intuitive to preserve last sketch vertex (add it to the feature before finishing drawing) or to preserve only the 'stable' ones (ones that are added by left click).
> 
> In the PR the first option is used. On the other hand tester who tested it on v1.24 asked for the second one.

Make it configurable :)

components/draw/draw.service.ts
```diff
@@ -375,6 +423,41 @@ export class HsDrawService {
         source: this.source,
         type: /** @type {GeometryType} */ this.type,
         style: changeStyle ? changeStyle() : undefined,
+        condition: (e) => {
```
DailisLangovskis, line 426: Should check if user tries to right click without adding any points on the map, otherwise getting null value.

DailisLangovskis:
Drawn features are not showing either on the map, or legend. This should have been fixed by #1066 .

components/draw/draw.service.ts
FilipLeitner: Not sure I understand what do you mean.

FilipLeitner:
> Drawn features are not showing either on the map, or legend. This should have been fixed by #1066 .

Could you be more specific please. Not observing such a behvior
image

components/draw/draw.service.ts
DailisLangovskis: When I select line or polygon, and click right mouse click on the map, I get error, cannot read length of undefined.

DailisLangovskis:
> > Drawn features are not showing either on the map, or legend. This should have been fixed by #1066 .
> 
> Could you be more specific please. Not observing such a behvior
> image

Yes, this was on my side, because I can replicate it again :D

components/draw/draw.service.ts
DailisLangovskis: https://drive.google.com/file/d/1-g6Kb_CRPaqOcBa6lDvIpczOVP2I9idM/view This is the error I was experiencing.

FilipLeitner: Oh I see, right . Thanks.

DailisLangovskis (approved):
Everything seems to be ok now!

components/draw/draw.service.ts
```diff
@@ -171,7 +184,7 @@ export class HsDrawService {
         visible: true,
         removable: true,
         editable: true,
-        synchronize: true,
```
raitisbe, line 174: What if we set this to true if and only if we have layman endpoint defined?

components/draw/draw.service.ts
```diff
@@ -335,12 +348,47 @@ export class HsDrawService {
...
+      HsConfirmDialogComponent,
+      {
+        message: this.TranslateService.instant('Really delete this layer?'),
+        title: this.TranslateService.instant('Confirm delete'),
+      }
+    );
+    const confirmed = await dialog.waitResult();
+    if (confirmed == 'yes') {
+      this.HsMapService.map.removeLayer(this.selectedLayer);
+      if (this.selectedLayer.get('synchronize') == true) {
```
raitisbe, line 367: Would be better to move insides of this if condition to layman service to seperate concerns. Drawing service shouldnt manage layers and do synchronization to endpoints.

components/draw/draw.service.ts
```diff
@@ -375,6 +423,41 @@ export class HsDrawService {
...
+          if (e.originalEvent.buttons === 1) {
+            //left click
+            return true;
+          } else if (e.originalEvent.buttons === 2) {
+            //right click
+            const minPoints = this.HsConfig.preserveLastSketchPoint ? 1 : 0;
+            if (this.type == 'Polygon') {
+              const vertexCount = this.draw.sketchLineCoords_?.length;
+              if (vertexCount >= 4 - minPoints) {
+                setTimeout(() => {
```
raitisbe, line 436: Try refactoring the duplicate 
```
 setTimeout(() => {
                  if (minPoints == 0) {
                    this.removeLastPoint();
                  }
                  this.draw.finishDrawing();
                }, 250);
``` 
into seperate function so it's more readable.

raitisbe (changes requested):

components/draw/draw.service.ts
FilipLeitner: 29c25580f13c56e014bb0b95065009ba68ec8403

components/draw/draw.service.ts
FilipLeitner: 15564b091777044edb8d270dce08ca85724d0e90

components/draw/draw.service.ts
FilipLeitner: 7cc8eaf00c19caaad4a9bb9d5643cb06ffa1d422

raitisbe (approved):

PR #1105, opened 2020-10-04
Added missing loadash dependency after initial install

DailisLangovskis:
Added this, because after clean pc windows install, this was the first error message, showing the dependency as missing.


raitisbe:
Could be that it came with some other of our dependencies but was removed from that. In hslayers we have lodash only in sortByPipe.class.ts. Alternative would be to rewrite it without lodash and drop the dependency.

raitisbe (approved):

PR #1109, opened 2020-10-05
Changed sidebar button sorting pipe logic 

DailisLangovskis:


jmacura:
When you are touching this file, I propose to rename `sortByPipe.class.ts` to `sort-by.pipe.ts`. For consistency.

PR #1119, opened 2020-10-07
Error for whiteboard composition with layer style loading 

raitisbe:
The problem with map integration was that we have in composition
```
'layers': [
          {
            'metadata': {},
            'visibility': true,
            'opacity': 1,
            'title': 'Measurement sketches',
            'className': 'Vector',
            'features': '{"type":"FeatureCollection","features":[]}',
            'maxResolution': null,
            'minResolution': 0,
            'projection': 'epsg:4326',
            'style': {
              'fill': 'rgba(255, 255, 255, 0.2)',
              'stroke': {'color': '#ffcc33', 'width': 2},
            },
          },
```
and the style property is being passed to OL layer constructor as is instead of being parsed and converted to openlayers Style object resulting in 
```
ERROR Error: Uncaught (in promise): AssertionError: Assertion failed. See https://openlayers.org/en/v6.4.3/doc/errors/#41 for details.
AssertionError: Assertion failed. See https://openlayers.org/en/v6.4.3/doc/errors/#41 for details.
```
and stop composition loading in its tracks.
This PR aims to fix that.

### DIFF
--- a/components/add-layers/vector/add-layers-vector.spec.ts
+++ b/components/add-layers/vector/add-layers-vector.spec.ts
@@ -8,9 +8,7 @@ import Map from 'ol/Map';
 import * as angular from 'angular';
 import {HsLayerUtilsService} from '../../utils/layer-utils.service';
 import {HsLayoutService} from '../../layout/layout.service';
-import {HsMapService} from '../../map/map.service';
-import {HsUtilsService} from '../../utils/utils.service';
-import { HsStylerService } from '../../styles/styler.service';
+import {HsStylerService} from '../../styles/styler.service';
 describe('add-layers-vector', () => {
   let el, scope, vm;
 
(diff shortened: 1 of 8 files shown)


components/add-layers/vector/add-layers-vector.spec.ts
```diff
@@ -34,6 +35,10 @@ describe('add-layers-vector', () => {
         }
       });
 
+      angular.module('hs.styles', []).service('HsStylerService', function () {
+        this.parseStyle = (new HsStylerService(null)).parseStyle;
```
jmacura: Strange way of doing things 😮

jmacura (approved):

components/add-layers/vector/add-layers-vector.spec.ts
raitisbe: I know, but it would no yield otherwise.

PR #1138, opened 2020-10-10
Map gallery update

FilipLeitner:
- Made base map inside of layermanager default again
- Added possibility to make basemap grayscale into layermanager
- Integrated layer-editor into basemap gallery so we have all potions avaliable for both basemap list types

### DIFF
--- a/components/layermanager/dimensions/layer-editor-dimensions.component.ts
+++ b/components/layermanager/dimensions/layer-editor-dimensions.component.ts
@@ -1,4 +1,5 @@
 import {Component, Input} from '@angular/core';
+import {HsConfig} from '../../../config.service';
 import {HsDimensionDescriptor} from './dimension.class';
 import {HsDimensionService} from '../../../common/dimension.service';
 import {HsEventBusService} from '../../core/event-bus.service';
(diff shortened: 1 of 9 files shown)


components/layermanager/layermanager-gallery.component.ts
```diff
@@ -58,4 +61,18 @@ export class HsLayerManagerGalleryComponent {
...
   }
+  expandMenu(e, layer) {
+    const parent = (e.target as Element).parentElement;
+    this.HsLayerManagerService.toggleLayerEditor(
+      layer,
+      'settings',
+      'sublayers'
+    );
+
+    if (parent.classList.contains('expanded')) {
```
raitisbe: can't we do this with ngClass instead? Is HsLayerManagerService.menuExpanded the same as HsLayerManagerGalleryComponent.menuExpanded and could we just use the variable in component since it deals mostly only with UI state.

components/layermanager/dimensions/layer-editor-dimensions.component.ts
```diff
@@ -13,13 +14,16 @@ import {Layer} from 'ol/layer';
 export class HsLayerEditorDimensionsComponent {
   @Input('ol-layer') olLayer: Layer;
   dimensions: Array<HsDimensionDescriptor> = [];
-
+  gallery: boolean;
```
raitisbe: galleryEnabled maybe?

raitisbe (changes requested):

components/layermanager/dimensions/layer-editor-dimensions.component.ts
FilipLeitner: Got there by accident . Deleted here 389232684772b6872930dc71cf1a317b07b920ca

components/layermanager/layermanager-gallery.component.ts
FilipLeitner: Its there :) 
https://github.com/hslayers/hslayers-ng/pull/1138/commits/476d99605d9420727519f9b0a84970beca384f99#diff-a616dc4693b080d259bac93af2291d92R72

PR #1155, opened 2020-10-19
Translate sensors

raitisbe:


components/language/language.service.ts
```diff
@@ -71,7 +71,15 @@ export class HsLanguageService {
    * @param params
    * @returns {string} Translation
    */
-  getTranslation(str: string, params?: object): string {
+  getTranslation(str: string, params?: any): string {
     return this.TranslateService.instant(str, params);
   }
+
+  getTranslationIgnoreNonexisting(module: string, text: string): string {
```
jmacura, line 78: `getTranslationIgnoreNonExisting` with capitalized E would be nicer 😉

components/language/language.service.ts
```diff
@@ -71,7 +71,15 @@ export class HsLanguageService {
...
+    const tmp = this.getTranslation(module + '.' + text);
+    if (tmp.indexOf(module + '.') > -1) {
```
jmacura, line 80: I personally prefer `.includes(module + '.')` over `indexOf(module + '.') > -1` – it's shorter and intentions are a bit clearer 😉

raitisbe: Indeed

raitisbe:
https://github.com/hslayers/hslayers-ng/pull/1155/commits/6d4126d78a7a513b551378b031dbff4bea5b06b6 might be interesting for you @FilipLeitner and @DailisLangovskis

components/core/core.module.ts
```diff
@@ -1,3 +1,4 @@
+import * as merge from 'deepmerge';
```
jmacura, line 1: We have [structuredClone()](https://github.com/hslayers/hslayers-ng/blob/32cd15abeb92ee121c831c24e192c6b31d1570b2/components/utils/utils.service.ts#L329) in our UtilsService, in case it can be helpful

raitisbe: Thanks, I will try. Would be good to leave out a dependency

FilipLeitner (approved):

components/core/core.module.ts
raitisbe: On the second hand, it's just 700bytes compressed and merge.all does more than our utils function and more elegantly. I think I'm gonna keep it after all :)

jmacura: @raitisbe If it is more powerful and we will included it in out dependencies, then we can probably remove our custom function and replace its occurances with the deepmerge version - for consistency and easier debugging. What do you think?

raitisbe: @jmacura Was afraid to break something and was in a rush, but I agree.

PR #1188, opened 2020-10-26
Wfs bbox

FilipLeitner:
Allow adding WFS layers from Micka


jmacura:
Ported to 2.x in #1224

PR #1197, opened 2020-10-28
Added translations to rest of compositions html

DailisLangovskis:
Note: Added translation strings also to cs.json. but did not translate them. Would like @FilipLeitner or @jmacura  to do this, so there is no grammar mistakes.
P.s. For anyone, who is adding translations, please add them also to other local language (lv or cs) in our case, and tell that translated text is missing for the language rather then just don't add anything at all. It is than easier to fix it the going through json files and looking for differences.


raitisbe:
Thanks. I think I'm going to merge now, but please keep adding the commits. Better to merge these things as soon as possible since html conflicts are kinda hard to resolve

jmacura:
Thank you very much @DailisLangovskis ! I will gladly go through the cs non/translations and add/correct them 🙂 This branch seems like a good place to put them.

PR #1206, opened 2020-10-29
Fix layman sync problems

raitisbe:
Multiple problems are fixed here:
* Layer description (wfs, wms endpoint urls) were first retrieved for anonymous (browser) user, cached and on subsequent requests wrongly used even after user authorizes resulting in:
"https://groundwater.smartagro.lv/layman/client/geoserver/browser/wfs"
where it should have been 
"https://groundwater.smartagro.lv/layman/client/geoserver/test/wfs"
* Too many current-user requests to layman get issued for each synchronized layer
* 'events-suspended' attribute could be set in uick succession multiple in async function, hich could result in addfeature/changefeature events executed when they are actually pulled from layman, not drawn by user.

### DIFF
--- a/components/layermanager/layer-editor-vector-layer.service.ts
+++ b/components/layermanager/layer-editor-vector-layer.service.ts
@@ -99,7 +99,7 @@ export class HsLayerEditorVectorLayerService {
     });
   }
   updateFeatureTableLayers(layer: Layer): void {
-    const currentLayerIndex = this.HsConfig.layersInFeatureTable.findIndex(
+    const currentLayerIndex = this.HsConfig.layersInFeatureTable?.findIndex(
       (l) => l == layer
     );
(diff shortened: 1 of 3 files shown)


FilipLeitner (approved):

components/layermanager/layer-editor-vector-layer.service.ts
```diff
@@ -99,7 +99,7 @@ export class HsLayerEditorVectorLayerService {
     });
   }
   updateFeatureTableLayers(layer: Layer): void {
-    const currentLayerIndex = this.HsConfig.layersInFeatureTable.findIndex(
+    const currentLayerIndex = this.HsConfig.layersInFeatureTable?.findIndex(
```
raitisbe, line 102: Thanks

PR #1216, opened 2020-11-03
1.25 fixes

FilipLeitner:
- clustering
- wms layer max resolution
- Add support for .wmc compositions
- Add compositions error dialog
- sublayers styling react to resolution changes

### DIFF
--- a/components/layermanager/layermanager-metadata.service.js
+++ b/components/layermanager/layermanager-metadata.service.js
@@ -183,7 +183,9 @@ export default function (
               layer.getProperties()
             );
             console.log('final', maxScale);
-            layer.set('maxResolution', me.roundToHundrets(maxScale));
+            if(maxScale){
+              layer.set('maxResolution', me.roundToHundrets(maxScale));
+            }
           });
           return true;
         })
--- a/components/layermanager/layermanager-metadata.service.js
+++ b/components/layermanager/layermanager-metadata.service.js
@@ -62,6 +62,33 @@ export default function (
       }
     }
   };
+
+  /**
+   * Round number to hundrets
+   *
+   * @param {number} num
+   * @returns {number}
+   */
+  me.roundToHundrets = function (num) {
+    return Math.ceil(num / 100) * 100;
+  };
+
+  me.searchForScaleDenominator = function (layer) {
+    let maxScale = layer.MaxScaleDenominator ? layer.MaxScaleDenominator : null;
+    const layers = layer.Layer;
+    for (const index in layers) {
+      if (
+        layers[index].MaxScaleDenominator &&
+        maxScale < layers[index].MaxScaleDenominator
+      ) {
+        maxScale = layers[index].MaxScaleDenominator;
+      } else if (layers[index].Layer) {
+        const subScale = me.searchForScaleDenominator(layers[index]);
+        maxScale = subScale > maxScale ? subScale : maxScale;
+      }
+    }
+    return maxScale;
+  };
   /**
    * @function queryMetadata
    * @memberOf HsLayermanagerMetadata.service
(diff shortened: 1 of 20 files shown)


components/layermanager/layermanager-metadata.service.js
```diff
@@ -183,7 +183,9 @@ export default function (
               layer.getProperties()
             );
             console.log('final', maxScale);
-            layer.set('maxResolution', me.roundToHundrets(maxScale));
+            if(maxScale){
```
raitisbe, line 186: Hundreds :)

components/layermanager/layermanager-metadata.service.js
```diff
@@ -62,6 +62,33 @@ export default function (
...
+   * @returns {number}
+   */
+  me.roundToHundrets = function (num) {
+    return Math.ceil(num / 100) * 100;
+  };
+
+  me.searchForScaleDenominator = function (layer) {
+    let maxScale = layer.MaxScaleDenominator ? layer.MaxScaleDenominator : null;
+    const layers = layer.Layer;
+    for (const index in layers) {
```
raitisbe, line 79: for of  better maybe?

components/layermanager/layermanager-metadata.service.js
```diff
@@ -143,7 +170,23 @@ export default function (
...
+          $timeout(() => {
+            if (layer.get('MaxScaleDenominator')) {
+              layer.set(
+                'maxResolution',
+                me.roundToHundrets(layer.get('MaxScaleDenominator'))
+              );
+              return;
+            }
+            const maxScale = me.searchForScaleDenominator(
+              layer.getProperties()
```
raitisbe, line 183: Why not just layer without getProperties?

components/layermanager/layermanager.service.js
```diff
@@ -713,9 +711,12 @@ export default function (
    * @function isLayerInResolutionInterval
    * @memberOf HsLayermanagerService
    * @param {Ol.layer} lyr Selected layer
-   * @description Test if layer (WMS) resolution is within map resolution interval
+   * @description Test if layer (WMS) resolution is NOT within map resolution interval. True = out of the interval
```
raitisbe, line 714: I think it should have been "Map resolution is within layers resolution interval minResolution <= current <= maxResolution"

raitisbe: the condition
```
lyr.getMinResolution() >= cur_res && cur_res <= lyr.getMaxResolution()
```
Seems to be the correct one.

components/layermanager/layermanager-metadata.service.js
FilipLeitner: Not sure why I went this way. ty

FilipLeitner: searchForScaleDenominator is recursive while sublayers are not instances of ol.layer so they dont have getProperties method

components/layermanager/layermanager.service.js
FilipLeitner: This ones bit confusing. The logic while using the return vaule is twisted:
`grayed: me.isLayerInResolutionInterval(layer)` .. should not be grayed if in resolution

FilipLeitner: Made some changes, seems to work fine now.
2a93397e3017698934abee272a92e293aee594a9

components/layermanager/layermanager.service.js
```diff
@@ -730,6 +733,7 @@ export default function (
       cur_res = HsMapService.map.getView().getResolution();
     }
     me.currentResolution = cur_res;
+    console.log(cur_res, lyr.getMaxResolution());
```
raitisbe, line 736: Probably a leftover?

FilipLeitner: O right, thanks. Will resolve with backporting of https://github.com/hslayers/hslayers-ng/pull/1206

PR #1219, opened 2020-11-04
Added customFeatureAction for clicked feature

DailisLangovskis:
Have created a way of calling a custom action on single feature clicked on the Hslayers map. To use this functionality, a layer must be defined inside app.js and layer attribute "onFeatureSelected" must be set to be a custom function, shown in the example.png.
example com

### DIFF
--- a/components/query/query-base.service.js
+++ b/components/query/query-base.service.js
@@ -1,3 +1,4 @@
+/* eslint-disable angular/on-watch */
 import Feature from 'ol/Feature';
 import VectorLayer from 'ol/layer/Vector';
 import {Circle, Fill, Stroke, Style} from 'ol/style';
@@ -62,7 +63,6 @@ export default function (
   this.featuresUnderMouse = [];
   this.featureLayersUnderMouse = [];
   this.dataCleared = true;
-
   /**
    *
    */
@@ -90,7 +90,6 @@ export default function (
       me.data.selectedProj = me.data.coordinates[0].projections[0];
       $rootScope.$broadcast('mapQueryStarted', evt);
     });
-
     if (
       angular.isDefined(HsConfig.popUpDisplay) &&
       HsConfig.popUpDisplay === 'hover'
@@ -107,16 +106,9 @@ export default function (
         'singleclick',
         HsUtilsService.debounce(me.showPopUp, 500, false, me)
       );
-    } /* else none */
-  }
-
-  /**
-   * @param e Event, which triggered this function
-   */
-  this.showPopUp = function (e) {
-    if (e.dragging) {
-      return;
     }
+  }
+  me.getFeaturesUnderMouse = function (e) {
     const map = e.map;
     me.featuresUnderMouse = map
       .getFeaturesAtPixel(e.pixel)
@@ -158,7 +150,15 @@ export default function (
       me.featuresUnderMouse = [];
     }
   };
-
+  /**
+   * @param e Event, which triggered this function
+   */
+  this.showPopUp = function (e) {
+    if (e.dragging) {
+      return;
+    }
+    me.getFeaturesUnderMouse(e);
+  };
   /**
    * @param feature
    */
--- a/components/query/query-vector.service.js
+++ b/components/query/query-vector.service.js
@@ -103,10 +103,7 @@ export default function (
     if (angular.isDefined(feature)) {
       original = feature;
       if (original.get('features') && original.get('features').length == 1) {
-        original = original
-          .get('features')
-          .map((feature) => feature)
-          .pop();
+        original = original.get('features')[0];
       }
     } else {
       original = null;
--- a/components/query/query-vector.service.js
+++ b/components/query/query-vector.service.js
@@ -1,3 +1,4 @@
+/* eslint-disable angular/on-watch */
 import * as extent from 'ol/extent';
 import {Select} from 'ol/interaction';
 import {Vector as VectorSource} from 'ol/source';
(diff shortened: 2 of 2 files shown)


raitisbe:
Why do we need customFeatureSelection attribute?

components/query/query-base.service.js
```diff
@@ -99,6 +99,12 @@ export default function (
         'pointermove',
         HsUtilsService.debounce(me.showPopUp, 500, false, me)
       );
+    }
+    if (me.customFeatureSelectionEnabled()) {
+      map.on(
+        'singleclick',
```
raitisbe, line 105: Why not just subscribe to  this.HsEventBusService.vectorQueryFeatureSelection event?

DailisLangovskis: At first I started to use this for checking if customAction should be called at all, but to be fair I forgot about this event. Release 1.25 does not even have HsEventBus :D

DailisLangovskis:
No more HsConfig property needed.

raitisbe (approved):
Ok from mys side

jmacura:
I receive this error when I try it :-(
image

components/query/query-vector.service.js
```diff
@@ -82,6 +81,38 @@ export default function (
...
+    if (angular.isDefined(feature)) {
+      original = feature;
+      original.get('features')
+        ? original.get('features').length == 1
+          ? (original = original.get('features'))
+          : (original = null)
+        : null;
+      return original;
+    }
+  };
```
jmacura, line 115: This function returns undefined when `feature` is not defined. I have no idea when it can happen, but it might be the source of the bug I observe.
Also, when `feature === undefined`, then the function returns `void`, which is technically `undefined`. So there is no need to check for `if (angular.isDefined(feature))` before `return`. The `return` statement can go out of this block along with       `original = feature;` assignment.

Investigating a bit further I don't think I understand the logic of this function at all... But the returned `null `seems to be causing the error.

components/query/query-vector.service.js
```diff
@@ -82,6 +81,38 @@ export default function (
...
   });
+  $rootScope.$on(
+    'vectorQuery.featureSelected',
+    HsUtilsService.debounce(
+      (event, feature) => {
+        const layer = feature.getLayer(HsMapService.map);
+        if (angular.isDefined(layer) && layer.get('onFeatureSelected')) {
+          const originalFeature = me
+            .getSingleFeature(feature)
+            .map((feature) => feature)
```
jmacura, line 92: What is this `map()` actually useful for?

jmacura (changes requested):

components/query/query-vector.service.js
DailisLangovskis: Map is needed to create a clone of feature, because other way I was experiencing that missing feature from the cluster source. Of course this would not be needed, if I wouldn't use .pop(). .pop() takes the object from memory reference and 'pulls it out', which caused that styling error. If You do think that it should be removed, I will do it!

components/query/query-vector.service.js
DailisLangovskis: This code has already been reconstructed btw. Also  the logic is - if feature  is defined, original = feature; this is done so that if the layer is not clustered, code doesn't break. Next I check if feature.get('features'), that means, that the layer is clustered. Than I need to know, if the selected feature has only one clustered source feature, if so, I set original = feature and return it.
code

components/query/query-vector.service.js
DailisLangovskis: onFeatureSelected: function (feature) {
    if (feature.get('myDirectLink')) {
      window.open(feature.get('myDirectLink'), '_blank');
    }
  }, this is the code I used in examples/full/app.js for testing

jmacura:
>   onFeatureSelected: function (feature) {
    if (feature.get('myDirectLink')) {
      window.open(feature.get('myDirectLink'), '_blank');
    }
  }, this is the code I used in examples/full/app.js for testing

With which data are you testing it? When I try it with examples/full for VectorLayer created from [this GeoJSON](https://raw.githubusercontent.com/jmacura/p4a-website-map/master/data/Plan4All_Members.geojson), I get literally nothing.

components/query/query-vector.service.js
jmacura: Ah, interesting. In that case, I would write it in a way I suggest in the new comment.
FTR, the .map() does not create a copy of `feature` either, it copies the reference to the same object, see e.g. https://www.freecodecamp.org/news/how-to-clone-an-array-in-javascript-1d3183468f6a/

components/query/query-vector.service.js
```diff
@@ -82,6 +81,38 @@ export default function (
...
+  );
+  me.getSelectedFeature = function (feature) {
+    let original;
+    if (angular.isDefined(feature)) {
+      original = feature;
+      if (original.get('features') && original.get('features').length == 1) {
+        original = original
+          .get('features')
+          .map((feature) => feature)
+          .pop();
```
jmacura, line 109:
```suggestion
        original = original.get('features').slice(-1)[0];
```

It could be some millisecond faster...

components/query/query-vector.service.js
jmacura: The functionality remains a bit obscured to me, but I made the code work now, so... This is resolved :-)

components/query/query-vector.service.js
DailisLangovskis: Okey, will replace it like so!

raitisbe: Since you know that the length == 1 then original.get('features')[0] should be enough

jmacura: That is very correct assumption @raitisbe , how could we have missed that?

jmacura (approved):
Great work, thank you!
Do you feel like re-implementing this functionality in the 2.x code too? 😄

DailisLangovskis:
> Great work, thank you!
> Do you feel like re-implementing this functionality in the 2.x code too? 😄

No problem! :) At least, now I know stuff.

PR #1218, opened 2020-11-03
Upgrade add-layers to ng9

jmacura:
also closes #947 and closes #950

PR #1235, opened 2020-11-06
Dont sync layman wfs layers based on synchronize attribute

raitisbe:


raitisbe:
need to set the definition attribute on layer when the layer is added from add-layers panel as wfs. Will do after https://github.com/hslayers/hslayers-ng/pull/1218 is merged

PR #1241, opened 2020-11-09
Sync fix

FilipLeitner:
- add geom change event listener for features added to source after the layer is pulled from layman
- Updating the geometry is not possible without the change of  [gemoetryName](https://openlayers.org/en/latest/apidoc/module-ol_Feature-Feature.html#getGeometryName) and [id](https://openlayers.org/en/latest/apidoc/module-ol_Feature-Feature.html#getId) that needs to match the one in layman

will be also neccessary to port to 2.x

### DIFF
--- a/components/save-map/layer-synchronizer.service.js
+++ b/components/save-map/layer-synchronizer.service.js
@@ -1,4 +1,5 @@
 import {WFS} from 'ol/format';
+
 const debounceInterval = 1000;
 
 /**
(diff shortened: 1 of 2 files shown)


raitisbe:
Strange, i didnt have any problems with updating geometry. Is this something speciffic in the latest layman?

raitisbe:
I guess I was always reloading the page after drawing new features, so didnt notice

FilipLeitner:
> Strange, i didnt have any problems with updating geometry. Is this something speciffic in the latest layman?

Most likely yes. Wasnt looking into laymans code tho

raitisbe:
No it was lack in my code. Layman wasnt to blame

PR #1248, opened 2020-11-10
Layout ng9

raitisbe:


components/layout/layout.component.ts
```diff
@@ -41,12 +41,6 @@ export class HsLayoutComponent {
...
       innerElement: '.hs-map-container',
     });
-    if (HsConfig.sidebarPosition === 'left') {
-      this.HsLayoutService.contentWrapper.classList.add('flex-reverse');
-      this.HsLayoutService.sidebarRight = false;
-    } else if (HsConfig.sidebarPosition != 'invisible') {
-      HsConfig.sidebarPosition = 'right';
-    }
 
     if (this.HsConfig.importCss == undefined || this.HsConfig.importCss) {
```
FilipLeitner, line 45: What would be the use case when we dont want to import css?

components/layout/layout.service.ts
```diff
@@ -1,197 +1,191 @@
...
-        },
-        {
-          enabled: true,
-          order: 8,
-          title: 'Save composition',
-          description: 'Save content of map to composition',
-          name: 'saveMap',
-          directive: 'hs.save-map',
-          mdicon: 'save',
+  data = {
```
FilipLeitner, line 9: This object seems to be deprecated

components/layout/layout.component.ts
raitisbe: if hslayers breaks some styles in the container application

components/layout/layout.service.ts
raitisbe: True, it was in the material design template. Deleted

PR #1250, opened 2020-11-11
Fix issue 1249: Pop-up window disappearing

### LINKED ISSUE #1249: Pop-up window disappears too quickly to be useful
pop-up element is part of the map, so the 'pointermove' event is fired even while moving cursor over the pop-up window. This is replaced after 0.5 s with a new pop-up for another feature, so the user has only those 0.5 s to read the content.

The behaviour is undesired in general. The pop-up window should remain opened while the cursor is moving over it.

jmacura:
fixes #1249 

It would be great if we can release minor bugfix release 2.4.1 with this PR. Today 😆

### DIFF
--- a/components/query/query-base.service.ts
+++ b/components/query/query-base.service.ts
@@ -126,7 +126,8 @@ export class HsQueryBaseService {
    * @param e Event, which triggered this function
    */
   showPopUp(e) {
-    if (e.dragging) {
+    // The latter case happens when hovering over the pop-up itself
+    if (e.dragging || e.originalEvent?.originalTarget?.tagName != 'CANVAS') {
       return;
     }
     const map = e.map;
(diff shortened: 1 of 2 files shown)


DailisLangovskis (approved):

raitisbe:
> Today

Do you realy need a release? Im lately compiling my applications against feature branches or develop locally and then publishing the bundle on server. Much easier than to create releases every day.

components/query/query-base.service.ts
```diff
@@ -126,7 +126,8 @@ export class HsQueryBaseService {
    * @param e Event, which triggered this function
    */
   showPopUp(e) {
-    if (e.dragging) {
+    // The latter case happens when hovering over the pop-up itself
+    if (e.dragging || e.originalEvent?.originalTarget?.tagName != 'CANVAS') {
```
raitisbe, line 130: Good idea! I had the same problem but couldnt think of a way how to solve it.

jmacura:
Fixed in 2.4.1 (this PR) and 2.5/3.0 (#1255).
**Not** back-ported to 1.27.x.

PR #1256, opened 2020-11-12
Sass integration

FilipLeitner:
- get rid of isolate-bootstrap 
- possiblity to theme apps using external 'custom.scss' file by modifying scss variables (ability to use these inside ts file to come)

For "full' apps the overrides file located in root of the project will be imported something like:
```
      {
        test: /\.scss$/,
        use: [
          'style-loader',
          'css-loader',
          {
            loader: 'sass-loader',
            options: {
              additionalData: fs.existsSync('./custom.scss')
                ? `@use "custom.scss" as *;`
                : '',
            },
          },
        ],
      },
```

For "simple" apps I tried simillar setup as for hub.lesprojekt
```
hsl-ng-2.x
 -hslayers-ng
     ...
     -css
        -app.scss
 -apps
   -simple
      -app.js
      -custom.css
```

and this config worked
...const scssOverridesPath = '../apps/simple/';
```
      {
        test: /\.scss$/,
        use: [
          'style-loader',
          'css-loader',
          {
            loader: 'sass-loader',
            options: {
              additionalData: fs.existsSync(scssOverridesPath + 'custom.scss')
                ? `@use "${scssOverridesPath}custom.scss" as *;`
                : '',
            },
          },
        ],
      },
```
Its bit supprising to me that It needs to be set only one folder up a tree '../' not '../../' as i expected it to be relative to app.scss file. Reason for that is that 'additionalData' param' does this:
image

If there are any other/better suggestions/solutions  for 'simple' apps let me know


jmacura (commented):
I can't see any difference amongst the changes in webpack.*.js files... Are there any? Why not putting it in webpack.common.js?

jmacura:
> Its bit supprising to me that It needs to be set only one folder up a tree '../' not '../../' as i expected it to be relative to app.scss file. Reason for that is that 'additionalData' param' does this:
> image
> 

As you are using `fs` to read the file, this rule shall apply: https://nodejs.org/api/fs.html#fs_file_paths

FilipLeitner:
> I can't see any difference amongst the changes in webpack.*.js files... Are there any? Why not putting it in webpack.common.js?

Ye, thanks for pointing that out. Well I got bit confused a it s the same with css loader, right? So I tought it has got to do something with the minification we do in prod/bundle. Not exactly sure how merge work but if its ok with those processes we might as well move both of the loaders to common (thinking about it now, it should be as it works with ts and js files).

Theres also another possible change...do we need `allowlist: ['bootstrap']` for nodeExternals as bootstrap is included within our scss?

raitisbe (approved):

PR #1258, opened 2020-11-12
Serialize style fix

DailisLangovskis:
Main improvement - Made functionality for improved existing layer style check. In cases when layer style has icon image, decode it and extract style properties, that will be reapplied in Styler panel.
Created for projects that loads map compositions with styled layers.

### DIFF
--- a/components/styles/styler.component.ts
+++ b/components/styles/styler.component.ts
@@ -244,7 +244,6 @@ export class HsStylerComponent {
   setStyleByJson(style_json: StyleJson): void {
     const style = new Style(style_json);
     const layer = this.HsStylerService.layer;
-    // const isClustered = this.HsLayerUtilsService.isLayerClustered(layer);
     switch (this.level) {
       case 'feature':
         this.setStyleForFeatures(layer, style);
@@ -294,7 +293,6 @@ export class HsStylerComponent {
    * @param {StyleLike|null} style Style to set for the feature. Can be null
    */
   private setStyleForFeatures(layer: VectorLayer, style: Style | null): void {
-    // const isClustered = this.HsLayerUtilsService.isLayerClustered(layer);
     const underlyingSource = this.HsStylerService.getLayerSource(layer);
     /**
      * We set a blank VectorSource temporarily
@@ -360,7 +358,8 @@ export class HsStylerComponent {
         path.style.strokeWidth = this.iconlinewidth;
       }
       this.serialized_icon =
-        'data:image/svg+xml;base64,' + window.btoa(iconPreview.innerHTML);
+        'data:image/svg+xml;base64,' + this.encodeTob64(iconPreview.innerHTML);
+      // window.btoa();
     }
   }
   /**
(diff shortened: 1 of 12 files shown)


DailisLangovskis:
New way of encoding Icon data was necessary for reverse engineered decoding.

components/styles/styler.component.ts
```diff
@@ -402,17 +417,13 @@ export class HsStylerComponent {
...
   }
-
   readCurrentStyle(layer: VectorLayer): void {
-    let style = layer.getStyle();
-    if (this.HsLayerUtilsService.isLayerClustered(layer)) {
-      style = layer.get('hsOriginalStyle');
-    }
-    if (typeof style == 'function') {
-      const resolvedStyle: Style | Style[] = style(new Feature());
+    const resolvedStyle = this.HsStylerService.getLayerStyleObject(layer);
```
raitisbe, line 421: Will this work even if we dont have features currently in the layer or we don't we are not interested in that since hasLine/poly/point will have false anyways?

components/styles/styler.component.ts
```diff
@@ -421,36 +432,95 @@ export class HsStylerComponent {
...
-    } else if (this.HsUtilsService.instOf(resolvedStyle, Style)) {
+    } else {
       this.parseStyle(resolvedStyle);
     }
   }
 
-  private parseStyle(subStyle: Style) {
-    if (subStyle.getStroke()?.getColor()) {
+  private parseStyle(style: Style) {
+    const subStyle: Style = this.HsSaveMapService.serializeStyle(style);
```
raitisbe, line 441: Why do we need to serialize first?

components/styles/styler.component.ts
```diff
@@ -421,36 +432,95 @@ export class HsStylerComponent {
...
+      if (image.stroke?.color) {
         this.iconlinecolor = this.HsStylerColorService.findAndParseColor(
-          subStyle.getStroke().getColor()
+          image.stroke.color
         );
       }
-      if (subStyle.image.getFill()?.getColor()) {
+    }
+  }
+  getIconStyleFromLayer(decodedIcon: any): void {
```
raitisbe, line 499: Would like to move these function as much as possible to service.

raitisbe:
> Main improvement - Made functionality for improved existing layer style check. In cases when layer style has icon image, decode it and extract style properties, that will be reapplied in Styler panel.
> Created for projects that loads map compositions with styled layers.

Thanks, it's definitely useful

components/styles/styler.component.ts
DailisLangovskis: Will do!

components/styles/styler.component.ts
DailisLangovskis: This serializeStyle function does lot of checks, which therefore are not needed to repeat, and also code looks more readable I think.

components/styles/styler.component.ts
DailisLangovskis: If should work, because the getLayerStyleObject, gets layer style if exists, and if style is function if creates an empty new feature, for it to work.

components/styles/styler.component.ts
```diff
@@ -49,9 +48,6 @@ export class HsStylerComponent {
   iconfillcolor: any;
   iconlinecolor: any;
   serialized_icon: any;
-  // hasLine: any;
```
DailisLangovskis: Confirmed by @raitisbe

PR #1264, opened 2020-11-14
No angularjs

raitisbe:


jmacura:
Awesome! I know it's tempting to merge this quickly and have things cleaner and faster, but I propose to create one more release in the 2.x line before merging this PR.
As removing AJS will be potentially a great breaking change for HSL-based applications, this could be then part of a version named 3.0, so we follow semver, at least particularly...

raitisbe:
Yes, I agree. I wanted also to get the aot compilation working and only after that merge to some temporary version-3-rc branch from where we can create the releases. 
Develop can wait

PR #1319, opened 2020-12-04
Issue 1318/add logo version

### LINKED ISSUE #1318: Add logo and version in UI
Add logo or name somewhere in the UI. Also expose current HSL version to the user. Good to get to know the version currently in use without checking sources. Proposed by Petr Vácha.

image

jmacura:
image

only displays when no panel is visible
image

fixes #1318


jmacura:
[Add logo and version in UI](https://trello.com/c/g9hwX7RB/77-add-logo-and-version-in-ui)

FilipLeitner (approved):
Looks good, we jsut might want to add some config toggle param in case we would like to hide it, for example if the map would be of smaller height.

jmacura:
> 
> 
> Looks good, we jsut might want to add some config toggle param in case we would like to hide it, for example if the map would be of smaller height.

In my opinion, our config is overwhelmed with options already. I was rather thinking about automatically and smoothly replacing a logo with a plain text in case the calculated height of the container element is smaller then the calculated img height. But that is too advanced styling for me 😄

FilipLeitner:
> > Looks good, we jsut might want to add some config toggle param in case we would like to hide it, for example if the map would be of smaller height.
> 
> In my opinion, our config is overwhelmed with options already. I was rather thinking about automatically and smoothly replacing a logo with a plain text in case the calculated height of the container element is smaller then the calculated img height. But that is too advanced styling for me 😄

Well yea you are right, we indeed have a lot of config values. Like your idea (Y) lets make an issue so its not forbiden and I can reach to it later

raitisbe:
Ported to 3x in https://github.com/hslayers/hslayers-ng/commit/c3c5da48bbfc67350124f5cee335b0cd53f6f717

raitisbe:
And this https://github.com/hslayers/hslayers-ng/commit/ea7c125dc4ded9debe5fa4129963945c2ea4e364 too. The problem was that we cant import package.json because its exists outside src directory - getting File  is not under 'rootDir' . 'rootDir' is expected to contain all source files... But a workaround is to copy the package.json to src on prebuild step and we can even strip the unneeded parts out ouf it that way.

jmacura:
Wow, so many new stuff and tricks to learn in 3.x ...

jmacura:
Backported to 1.x in #1369

Issue #1326, opened 2020-12-07
Keep layer title black when turned off


fzadrazil:
Layer title in layer manager should be in black color even when the layer is turned off. Grey out the title only for the case when the layer is outside the visible scale range.

fzadrazil:
![](https://github.trello.services/images/mini-trello-icon.png) [Layer title in black color when in the visible scale range](https://trello.com/c/NLiJiE3B/136-layer-title-in-black-color-when-in-the-visible-scale-range)

FilipLeitner:
Actually I kinda link it the way it is now: but ok if thats the way we want it :)

PR #1328, opened 2020-12-07
Patch 1.27

FilipLeitner:
#1326 Keep layer title black when turned off
Fix layerorder of layers loaded from permalink
Remove deprecated fromcrs param


FilipLeitner:
[Layer title in black color when in the visible scale range](https://trello.com/c/NLiJiE3B/136-layer-title-in-black-color-when-in-the-visible-scale-range)

PR #1340, opened 2020-12-10
Issue 383/file upload

DailisLangovskis:


components/add-layers/vector/add-layers-vector.service.ts
```diff
@@ -192,4 +199,49 @@ export class HsAddLayersVectorService {
       }
     }
   }
+  readUploadedFile(file: any): void {
+    let fileToJSON: any;
+    if (file.name.includes('.kml')) {
+      fileToJSON = kml(new DOMParser().parseFromString(file, 'text/xml'));
+      this.addNewLayer(fileToJSON);
+    } else if (file.name.includes('.gpx')) {
```
jmacura, line 206:
```suggestion
    } else if (file.name.endsWith('.gpx')) {
```

components/add-layers/vector/add-layers-vector.service.ts
```diff
@@ -192,4 +199,49 @@ export class HsAddLayersVectorService {
       }
     }
   }
+  readUploadedFile(file: any): void {
+    let fileToJSON: any;
+    if (file.name.includes('.kml')) {
```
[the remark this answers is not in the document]
DailisLangovskis: @jmacura Yes, thank You I am well aware of this! Just made it fast, so I can Test it! Now at least I don't have to look for the answer my self! :D

components/add-layers/vector/add-layers-vector.service.ts
```diff
@@ -192,4 +197,77 @@ export class HsAddLayersVectorService {
...
+      reader.onload = () => {
+        try {
+          const fileToJSON = JSON.parse(<string>reader.result);
+          if (fileToJSON !== undefined) {
+            if (fileToJSON.features.length > 0) {
+              this.createVectorObjectFromJson(fileToJSON);
+            }
+          }
+        } catch (e) {
+          // do nothing
```
raitisbe, line 216: Maybe some output to console could be usefull in case of error.

raitisbe (approved):

jmacura (commented):
The implementation is too eager. It adds the layer into the map right after the file is uploaded instead of after the click on "+ Add". So it adds the layer with some default title and once the user clicks the "+ Add", just an empty layer is created with no geometry.

raitisbe:
> The implementation is too eager. It adds the layer into the map right after the file is uploaded instead of after the click on "+ Add". So it adds the layer with some default title and once the user clicks the "+ Add", just an empty layer is created with no geometry.

I think it's not bad that the data is visualized immediately, but to make ui behave more or less in a similar way for all the layer types, we should probably just display some overview of the uploaded data. Number of features, maybe prefill the title of show/hide the "extract styles" checkbox for kml layers. That would mean that we also need an extra button to reset the panel state when the adding is complete

DailisLangovskis:
> > The implementation is too eager. It adds the layer into the map right after the file is uploaded instead of after the click on "+ Add". So it adds the layer with some default title and once the user clicks the "+ Add", just an empty layer is created with no geometry.
> 
> I think it's not bad that the data is visualized immediately, but to make ui behave more or less in a similar way for all the layer types, we should probably just display some overview of the uploaded data. Number of features, maybe prefill the title of show/hide the "extract styles" checkbox for kml layers. That would mean that we also need an extra button to reset the panel state when the adding is complete

In progress...

raitisbe (approved):

jmacura:
Well implemented I have to say.

PR #1347, opened 2020-12-11
Fix loader image in add-layers-shp

jmacura:
= properly resolve ES module


jmacura:
cherry-picked into #1353

PR #1348, opened 2020-12-11
Small fixes 2 5

FilipLeitner:


FilipLeitner:

config.service.ts
```diff
@@ -66,5 +66,6 @@ export class HsConfig {
   sizeMode: string;
   locationButtonVisible: any;
   openQueryPanelOnDrawEnd: boolean;
+  dss_disallow_add: boolean; //found in datasources-metadata-dialog
```
raitisbe, line 69: Wasnt this renamed to allowAddExternalDatasets ?

raitisbe: Never mind, that was singleDatasources. To be honest i dont see a reason why the list should be presented at all if the user isn't able to add the layer to map. Maybe lets drop this ugly config parameter?

raitisbe (approved):

PR #1353, opened 2020-12-14
Few 2.5 fixes

### LINKED ISSUE #1324: Redundant FROMCRS parameter causes WMS loading to fail
https://github.com/hslayers/hslayers-ng/blob/2a3d1f7ccb4d264d3986e80c6d82f015b111d097/components/compositions/compositions-parser.service.ts#L160

Is there a reason to have 'EPSG:3857' hardcoded here? In case of application with different projection, WMS fails to serve GetMap requests if the service does not support 3857.

### LINKED ISSUE #1349: Layman authorization opens in a new window
It should open in a dialog inside the map.

### LINKED ISSUE #1350: Remove only added layers with "Remove all layers" button
The "Remove all layers" icon in the Layer Manager should not remove layers which are defined in the app config file. In case we want it to be removable, the param can be set to true in the config.
Also, there are different labels. The icon is called "Remove all layers", but later in the dialog it says "Remove all added layers". This should be unified.

FilipLeitner:
Fix #1349 
Fix #1350
Partially fixes #1324. Havent deleted from cesium.


FilipLeitner:

jmacura:
superseded by #1362

PR #1356, opened 2020-12-14
Styler fixes for 2.5 release

DailisLangovskis:


DailisLangovskis:
> I was so deeply trying to implement WMS/WFS parsing from URL without introducing new Subject... yet it seems we gonna need it anyway (`layerSelectedFromUrl`).

Well this is used for styler panel only for now, but if any other use case is needed this can be redone for multiple purposes as well.

jmacura:

PR #1362, opened 2020-12-14
Fixes 1.25

### LINKED ISSUE #1324: Redundant FROMCRS parameter causes WMS loading to fail

### LINKED ISSUE #1349: Layman authorization opens in a new window

### LINKED ISSUE #1350: Remove only added layers with "Remove all layers" button

### LINKED ISSUE #1355: Layer editor opens at wrong place after the page is reloaded
https://localhost:8080/?hs_panel=layermanager&hs_x=1609802.342187116&hs_y=6396453.6057200935&hs_z=10&visible_layers=OpenStreetMap&layerSelected=WMTS

image

### LINKED ISSUE #1359: Composition layers overlay drawing
When loading some compositions from Micka and then drawing new features in the map, those are hidden behind the added layers. Tested with "Covid map 1st September 2020".
image

### LINKED ISSUE #1363: Unsaved drawing gets cleared when saved
After adding some features to the unsaved drawing and then trying to save it to a new layer, those featuers already drawn dissapear.

FilipLeitner:
Fix #1349 
Fix #1350
Partially fixes #1324. Havent deleted from cesium.
Fix #1355
Fix #1363
FIX #1359


components/layermanager/layermanager.service.ts
```diff
@@ -640,7 +640,7 @@ export class HsLayerManagerService {
   removeAllLayers(): void {
     const to_be_removed = [];
     this.HsMapService.map.getLayers().forEach((lyr) => {
-      if (lyr.get('removable') == undefined || lyr.get('removable') == true) {
+      if (lyr.get('removable') == true) {
```
raitisbe, line 643: Could you add a comment so we dont forget how its supposed to work?

raitisbe (approved):

components/layermanager/layermanager.service.ts
FilipLeitner: Yes, of course

assets/locales/lv.json
```diff
@@ -45,7 +45,8 @@
     "thisIsRequired": "Šis ir nepieciešams!",
     "tileSize": "\"Flīžu\" izmērs",
     "useProxy": "Izmantojiet starpniekserveri",
-    "useTiles": "Izmantojiet \"flīzes\""
+    "useTiles": "Izmantojiet \"flīzes\"",
+    "addSelected": "Add selected layers to the map"
```
jmacura, line 49: @DailisLangovskis @raitisbe Perhaps one of you could fill in the translation before merge?

components/add-layers/wms/add-wms-layer.directive.html
```diff
@@ -98,9 +98,11 @@
             </tbody>
         </table>
 
-        <button class="btn btn-primary" style="position: sticky;
-        bottom: 1.5rem; float: left;" (click)="selectAllLayers(data.services)">{{'ADDLAYERS.selectAllLayers' | translate}}</button>
-        <button class="btn btn-primary btn-sm" style="position: sticky;
-        bottom: 1.5rem; float: right;" (click)="addLayers(true)"><i class="icon-plus"></i></button>
+
+        <div class="w-100 d-flex justify-content-between bg-light pb-2" style="position: sticky;
```
jmacura, line 102: Very pretty, nicely done 🙂

jmacura (approved):
I have tested what I could and it all work, so if no more commits are coming, it's good to go.
Just note my note about translation 😉

assets/locales/lv.json
raitisbe: @jmacura done

PR #1370, opened 2020-12-14
Proxy wmts layers even after source change

raitisbe:
Port of https://github.com/hslayers/hslayers-ng/commit/dd908ef5abfa3525d770beb0347587fc8f0a2f65

### DIFF
--- a/components/map/map.service.ts
+++ b/components/map/map.service.ts
@@ -6,6 +6,7 @@ import Kinetic from 'ol/Kinetic';
 import Map from 'ol/Map';
 import Static from 'ol/source/ImageStatic';
 import View from 'ol/View';
+import WMTS from 'ol/source/WMTS';
 import proj4 from 'proj4';
 import {Cluster, OSM, Vector} from 'ol/source';
 import {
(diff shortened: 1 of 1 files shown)


components/map/map.service.ts
```diff
@@ -6,6 +6,7 @@ import Kinetic from 'ol/Kinetic';
 import Map from 'ol/Map';
 import Static from 'ol/source/ImageStatic';
 import View from 'ol/View';
+import WMTS from 'ol/source/WMTS';
```
jmacura, line 9: Move to line 11 as
`import {Cluster, OSM, Vector, WMTS} from 'ol/source';` ?
The same can be done with `ImageStatic as Static`

raitisbe: Done

components/map/map.service.ts
```diff
@@ -505,6 +500,45 @@ export class HsMapService {
...
+   * @param lyr {Layer} Layer which to proxify if needed
+   * @description Proxify layer based on its source object type and if its tiled or not.
+   * Each underlying OL source class has its own way to override imagery loading.
+   */
+  proxifyLayer(lyr: Layer): void {
+    const source = lyr.getSource();
+    if (
+      [ImageWMS, ImageArcGISRest].some((typ) =>
+        this.HsUtilsService.instOf(source, typ)
+      )
```
FilipLeitner, line 513:
```suggestion
      [ImageWMS, ImageArcGISRest].some((type) =>
        this.HsUtilsService.instOf(source, type)
      )
```

components/map/map.service.ts
```diff
@@ -505,6 +500,45 @@ export class HsMapService {
...
+    ) {
+      this.proxifyLayerLoader(lyr, false);
+    }
+    if (this.HsUtilsService.instOf(source, WMTS)) {
+      source.setTileLoadFunction((i, s) => this.simpleImageryProxy(i, s));
+    }
+    if (
+      [TileWMS, TileArcGISRest].some((typ) =>
+        this.HsUtilsService.instOf(source, typ)
+      )
```
FilipLeitner, line 523:
```suggestion
      [TileWMS, TileArcGISRest].some((type) =>
        this.HsUtilsService.instOf(source, type)
      )
```

FilipLeitner (changes requested):

components/map/map.service.ts
raitisbe: I was using it on purpose not to conflict with the type keyword.

FilipLeitner: Oh, I tought its just typo ....well but its just variable in a specific context, dont think it could be problematic in any way. However for the same reason I might also dont mind 'typ' then...it just kinda feels bit strange.

FilipLeitner (approved):

PR #1373, opened 2020-12-15
Drawings not visible

FilipLeitner:
Removing layer, which is set as "selectedLayer" for drawing, using "Remove all layers" button in layermanager doesnt remove its refference in drawing tool.  After the layer is added to the  map once again it get kinda duplicated and newly added drawings are not visible

### DIFF
--- a/components/draw/draw.service.ts
+++ b/components/draw/draw.service.ts
@@ -416,17 +416,16 @@ export class HsDrawService {
       .getLayers()
       .getArray()
       .filter((layer) => this.HsLayerUtilsService.isLayerDrawable(layer));
-
-    if (drawables.length == 0) {
-      //&& !this.selectedLayer
+    if (drawables.length == 0 && !this.tmpDrawLayer) {
       this.type = null;
       this.deactivateDrawing();
       this.selectedLayer = null;
-    } else {
+    } else if (drawables.length > 0) {
       if (
         !drawables.some(
-          (layer) => layer.get('title') == this.selectedLayer.get('title')
-        )
+          (layer) => layer.get('title') == this.selectedLayer?.get('title')
+        ) ||
+        !this.selectedLayer
       ) {
         this.selectedLayer = drawables[0];
         this.changeDrawSource();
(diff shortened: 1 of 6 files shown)


components/draw/draw.service.ts
```diff
@@ -263,7 +263,9 @@ export class HsDrawService {
    */
   addDrawLayer(layer: Layer): void {
     this.HsMapService.map.addLayer(layer);
-    this.fillDrawableLayers();
+    setTimeout(() => {
```
raitisbe, line 266: Why do we need timeout? Can you add a note please, so we dont remove it later?

FilipLeitner: Oh, most likely we dont. There was a problem with selectedLayer interaction inside fillDrawable layers. At first I thought it just need to wait a bit before refilling..turned out to be wrong condition. I just pushed to test on server and then forgot to delete and squash. Most likely because I thought it doesnt really matter in this specific case

components/draw/draw.service.ts
```diff
@@ -434,7 +434,7 @@ export class HsDrawService {
     const confirmed = await dialog.waitResult();
     if (confirmed == 'yes') {
       this.HsMapService.map.removeLayer(this.selectedLayer);
-      if (this.selectedLayer.get('synchronize') == true) {
+      if (this.selectedLayer.get('definition')?.format == 'hs.format.WFS') {
```
raitisbe, line 469: You can also use HsLayerSynchronizerService.isLayerSynchronizable(layer) for this

FilipLeitner: Id have to move it to the save map layman.service first, because of circular dependency. But I guess it might make a sense to be able to reuse the thin

components/draw/draw.service.ts
FilipLeitner: Had to redo the condition once more, wasnt quite working still.  https://github.com/hslayers/hslayers-ng/pull/1373/commits/d592b4fe28c7c289b7a73b7ed2a5322b6faeb79c

raitisbe (approved):

components/draw/draw.service.ts
raitisbe: I was afraid about that. Lets postpone moving it for version 3?

jmacura:

Issue #1236, opened 2020-11-08
SHP uploaded to Layman via HSL is not immediately added to the map


jmacura:

jmacura:
for the same request in 2.x see #702

jmacura:
Now obsolete and fixed in the consecutive major releases.

PR #1376, opened 2020-12-16
Add Shapefile automatically to map

### LINKED ISSUE #702: Add Shapefile automatically to map
When adding SHP, the layer is stored in Layman, but is not added automatically to the map.

jmacura:
fixes #702


jmacura:
Ugly one. I hate these Layman and Micka related bugs, they are so hard to catch... 🙁

raitisbe:
Maybe in that case it would be better to use the name returned from layman post request to make it futureproof? 
image
Did you notice post requests failing because of bad names?

components/add-layers/shp/add-layers-shp.component.ts
```diff
@@ -69,27 +72,30 @@ export class HsAddLayersShpComponent implements OnInit {
...
+    layerName: string
+  ): Promise<HsLaymanLayerDescriptor> {
+    try {
+      const descriptor = await this.hsLaymanService.describeLayer(
+        endpoint,
+        layerName
+      );
+      if (['STARTED', 'PENDING', 'SUCCESS'].includes(descriptor.wms.status)) {
+        setTimeout(async () => {
+          return await this.describeNewLayer(endpoint, layerName);
```
raitisbe: I was struggling with this timeout and thats the reason why I havent rewriotten to promises. It was returning inside setTimeout instead of describeNewLayer. I suspect it might be that here too. The way to check would be to upload a large shp file which takes time to process.

jmacura: You are commenting on an outdated file, are you aware of that? Exactly this part was the problem – I had to promisify the `setTimeout()` with `new Promise()` and resolve from inside of it.

jmacura:
> 
> Maybe in that case it would be better to use the name returned from layman post request to make it futureproof?

Definitely! That way we can go away with the unidecode as well.

jmacura:
> Did you notice post requests failing because of bad names?

I did not. Can you share more details about those failing requests?

jmacura:

raitisbe (approved):

jmacura:
For backport to 1.x refer #1236 , port to 3.x is needed.

PR #1378, opened 2020-12-16
Fixed some bugs

DailisLangovskis:


css/app.scss
```diff
@@ -546,6 +546,9 @@ button.but-title-sm {
 	position: relative;
 	font-family: Roboto, "Helvetica Neue", sans-serif;
 }
+.hs-ows-wms-capabilities-error {
```
jmacura, line 549: Not really a fix, but a workaround, right? The font-family shall be inherited from .hs-content-wrapper to all the dialogs, which is not..

jmacura:

css/app.scss
DailisLangovskis: Well, yeah, I don't know how to do it in other way, but at least made it, so it works for now.

jmacura (approved):
🐛 🔫

PR #1391, opened 2020-12-17
Dont re-add baselayers

### LINKED ISSUE #1390: Layers hidden behind basemap in shared map
Share map panel -> Pure map or Permalink. OSM basemap gets rendered over the rest of the map content.

FilipLeitner:
Prevents layers from getting under the base map on permalink layer load. Should be enough to fix #1390

### DIFF
--- a/components/map/map.service.ts
+++ b/components/map/map.service.ts
@@ -557,7 +557,9 @@ export class HsMapService {
     visibilityOverrides?: Array<string>
   ): Layer {
     if (removeIfExists && this.layerAlreadyExists(lyr)) {
-      this.removeDuplicate(lyr);
+      if (lyr.get('base') == true) {
+        return;
+      }
     }
     if (visibilityOverrides) {
       lyr.setVisible(this.layerTitleInArray(lyr, visibilityOverrides));


components/map/map.service.ts
```diff
@@ -557,7 +557,9 @@ export class HsMapService {
     visibilityOverrides?: Array<string>
   ): Layer {
     if (removeIfExists && this.layerAlreadyExists(lyr)) {
-      this.removeDuplicate(lyr);
+      if (lyr.get('base') == true) {
```
raitisbe, line 560: We wont be able to load base layers from compositions this way, no?

raitisbe: ok, I guess this.layerAlreadyExists will prevent them not being loaded, but why did they go under base map? I thought that layers are always added on top of the existing. Can't be that base layers are saved last to the permalink on top of the overlays? In that case, that's the cause of the issue, not removingDuplicate.

FilipLeitner: Problem is that layers that are added from app config are added first and only then the permalinks.

FilipLeitner: > We wont be able to load base layers from compositions this way, no?

Shouldnt be problem. Both conditions have to be met - removeIfExists  and layerAlreadyExists

raitisbe (approved):

PR #1399, opened 2020-12-18
Utils unit tests

DailisLangovskis:


components/utils/layer-utils.spec.ts
```diff
@@ -4,6 +4,7 @@ import 'core-js/es7/reflect';
 import 'reflect-metadata';
 import 'zone.js/dist/zone';
 import 'zone.js/dist/zone-testing';
+import OSM from 'ol/source/OSM';
 import Vector from 'ol/source/Vector';
 import VectorLayer from 'ol/layer/Vector';
```
jmacura, line 9: Could you squeeze the imports to `import {Cluster, ImageWMS, OSM, TileWMS, Vector, WMTS} from 'ol/source'`? Or expand them all into a single-line variant, but I don't like the mix..
Analogously with imports from `'ol/layer'`.

jmacura (commented):
Well done!
I have just a stylistic note.

raitisbe (approved):

PR #1400, opened 2020-12-18
Handle 404 in describeLayer properly

FilipLeitner:
- Not possible to upload new layer to layman.
When layer was not found on Layman, server responded with code 404 which needs to be handled inside catch block. Maybe some change on layman side after the update

### DIFF
--- a/components/save-map/layman.service.ts
+++ b/components/save-map/layman.service.ts
@@ -355,7 +355,7 @@ export class HsLaymanService implements HsSaverService {
         return response;
       }
     } catch (ex) {
-      if (ex.error.code == 15) {
+      if (ex.error?.code == 15) {
         return null;
       } else {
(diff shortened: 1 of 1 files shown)


components/save-map/layman.service.ts
```diff
@@ -350,15 +351,16 @@ export class HsLaymanService implements HsSaverService {
...
-      if (response?.code == 15) {
-        return null;
-      }
       if (response.name) {
         return response;
       }
     } catch (ex) {
-      this.HsLogService.error(ex);
-      throw ex;
+      if (ex.error.code == 15) {
```
raitisbe, line 358: ex.error?.code ?

raitisbe (approved):

FilipLeitner:
Thanks for the improvement

PR #1405, opened 2020-12-18
Vector layer options type

raitisbe:
This also solves that we are not able to add layman layers from datasources panel.
We used to have array ["WMS", "WFS"] there, but its overriden by the request for metadata from layman
because it returns type: "layman.layer" type attribute. Wasnt there in the previous layman version,
so we didnt notice


components/layers/hs.source.Wfs.js
```diff
@@ -100,6 +100,5 @@ export default function (options) {
     projection: options.projection,
     strategy: loadingstrategy.bbox,
   });
-  src.defOptions = options;
```
FilipLeitner, line 103: This change just reminded me that this file is not used anymore. Its all handled inside add.layers.wfs

FilipLeitner (approved):

components/layers/hs.source.Wfs.js
raitisbe: I was suspecting that. Should I delete it?

PR #1412, opened 2020-12-21
Re-insert existing but unsupported permalink layers

raitisbe:


FilipLeitner:
I already added support for WMTS to 1x so if wanted I can port . I just didnt want to keep adding news commits to 2.5 endlessly

DailisLangovskis (approved):

raitisbe:
> I already added support for WMTS to 1x so if wanted I can port . I just didnt want to keep adding news commits to 2.5 endlessly

Yes, lets publish the release how it is

PR #1461, opened 2021-01-05
Fixed missing logo

DailisLangovskis:


raitisbe:
Can't be that simple :)

DailisLangovskis:
> Can't be that simple :)
Well I think it is, because the [hidden] property logoDisabled was set to true.

DailisLangovskis:
> This is really a streamline fix of the issue I have reported 😄
> I can see you have made much nicer implementation in the `hide-hs-logo` branch, @DailisLangovskis, so why opening this PR?
Well, because that branch does not work and I had communicated with @FilipLeitner, that he will try to make the logo disappear with scss magic.

PR #1463, opened 2021-01-05
Re-enable doc script

jmacura:
This re-adds the @compodoc/compodoc as a devDependency and re-enables the "doc" script in package.json.

Shortcomings of current implementation: documentation only for main "hslayers" library is generated, not for -cesium or -sensors part.

Advantages of current implementation: we now have pretty DOM tree tabs in the docs 🎉 
image


fzadrazil (approved):

jmacura:
Wait with the merge a little while longer please, I am now trying to make the docs script create 3 documentations for all 3 libs.

PR #1465, opened 2021-01-06
Move 'layers' from components to common

jmacura:
This PR moves layers (of which currently only SparqlJson is in active use) to common folder.
Furthermore, it exposes SparqlJson in hslayersNgConfig() and in public-api.
Under the hood the SparqlJson was rewritten from function returning `ol/source/Vector` to a class extending the `ol/source/Vector`.


raitisbe:
Great job! Appreciate the import reorganizing.

raitisbe (approved):

PR #1468, opened 2021-01-08
Compositions rework

DailisLangovskis:
There are still a lot to do and many bugs and features to fix/improve.


DailisLangovskis:
In need of CS translations

projects/hslayers/src/components/compositions/compositions.html
```diff
@@ -34,7 +34,7 @@
                     name="hs-filter-by-name" [(ngModel)]="HsCompositionsCatalogueService.data.query.title"
                     (change)="loadCompositions()">
                 <div class="input-group-append">
-                    <button class="btn input-group-text text-secondary border-left-0" (click)="loadCompositions()">
+                    <button class="btn input-group-text text-secondary border-left-0" style="background-color: #f8f9fa" (click)="loadCompositions()">
```
raitisbe: I think it would be better not to hardcode colors, but use css classes defined in app.scss instead

DailisLangovskis: Will do that now!

jmacura:
Translating to cs done.

raitisbe (approved):

PR #1523, opened 2021-01-22
Emit layoutLoads event after subscriber registered

raitisbe:


jmacura:
Just a note: Another way to implement this without `ngAfterViewInit()` could be using [BehaviorSubject](https://rxjs-dev.firebaseapp.com/guide/subject#behaviorsubject).

raitisbe:
> Just a note: Another way to implement this without `ngAfterViewInit()` could be using [BehaviorSubject](https://rxjs-dev.firebaseapp.com/guide/subject#behaviorsubject).

Yes, I tried that at first, but then i have to provide some initial value i.e. null and in the subscriber check if it's != null. Seemed messy.

FilipLeitner (approved):

PR #1527, opened 2021-01-25
Do not show empty wms popup

DailisLangovskis:
Can You replicate the 404 popup with these changes?

### DIFF
--- a/projects/hslayers/src/components/query/query.component.ts
+++ b/projects/hslayers/src/components/query/query.component.ts
@@ -61,7 +61,10 @@ export class HsQueryComponent {
       this.HsQueryBaseService.getFeatureInfoCollected.subscribe(
         (coordinate) => {
           const invisiblePopup: any = this.HsQueryBaseService.getInvisiblePopup();
-          if (invisiblePopup.contentDocument.body.children.length > 0) {
+          const bodyElements = this.checkForBodyElements(
+            invisiblePopup.contentDocument.body.children
+          );
+          if (bodyElements?.length > 0) {
             //TODO: dont count style, title, meta towards length
             if (this.HsQueryBaseService.popupClassname.length > 0) {
               this.popup.getElement().className = this.HsQueryBaseService.popupClassname;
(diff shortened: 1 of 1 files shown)


projects/hslayers/src/components/query/query.component.ts
```diff
@@ -91,7 +94,18 @@ export class HsQueryComponent {
       );
     });
   }
-
+  checkForBodyElements(docChildren: any): any[] {
```
FilipLeitner, line 97: Wouldnt it be possible to use just getElementsByTagName() or querySelectorAll()

DailisLangovskis: Well getInvisiblePopup() already does that. Rest of the code checks if there is content for the popup available, otherwise empty popup won't be shown.

raitisbe: Could you use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some  on invisiblePopup.contentDocument.body.children array please? No need to have return the actual bodyElements list if we are just checking if any non-meta-head exists.

raitisbe (changes requested):

raitisbe (approved):

projects/hslayers/src/components/query/query.component.ts
```diff
@@ -91,7 +94,16 @@ export class HsQueryComponent {
       );
     });
   }
-
+  checkForBodyElements(docChildren: any): boolean {
```
raitisbe, line 97:
```suggestion
checkForBodyElements(docChildren: any): boolean {
    return docChildren.some((ch) => ch.tagName != 'META' && ch.tagName != 'STYLE'))
}
```

raitisbe (changes requested):

jmacura (approved):
I can not replicate the bug neither in develop or 1.27, so I am fine with this change.

raitisbe (approved):

PR #1532, opened 2021-01-25
Fix new layman layer pushing

raitisbe:
Basically just try catch blocks here, but sadly it shows like a lot of changes.

### DIFF
--- a/projects/hslayers/src/components/draw/draw.service.ts
+++ b/projects/hslayers/src/components/draw/draw.service.ts
@@ -167,13 +167,14 @@ export class HsDrawService {
     const layman = this.HsLaymanService.getLaymanEndpoint();
     const drawLayer = new VectorLayer({
       title: tmpTitle,
+      //TODO: Also name should be set, but take care in case a layer with that name already exists in layman
       source: tmpSource,
       show_in_manager: true,
       visible: true,
       removable: true,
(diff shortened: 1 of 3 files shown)


FilipLeitner (approved):
Also noticed this with new drawing layers, ty for fix. Some more problems might appear bacause of the auth via client/geoserver path but most of the fixes for that should be included in #1468. 
One thing that might be added to this PR is saving drawing layer name right at the creation: Currently only title is included ... its need when saving the new drawlayer to composition.
https://github.com/hslayers/hslayers-ng/blob/cf013cccc6406e454dd7e3714347c09f5aa6bc17/projects/hslayers/src/components/draw/draw.service.ts#L169

raitisbe:
> Also noticed this with new drawing layers, ty for fix. Some more problems might appear bacause of the auth via client/geoserver path but most of the fixes for that should be included in #1468.
> One thing that might be added to this PR is saving drawing layer 'name' param right at the creation: Currently only title is included ... its need when saving the new drawlayer to composition.
> https://github.com/hslayers/hslayers-ng/blob/cf013cccc6406e454dd7e3714347c09f5aa6bc17/projects/hslayers/src/components/draw/draw.service.ts#L169

I will better just add a todo comment. Have to explore more, how its handled when a layer with that name already exists

PR #1533, opened 2021-01-26
Add map sharing component to the hslayers-server

fzadrazil:
First version of the Status Manager replacement. I have been testing as a separate application so far.


raitisbe:
Small tweaks were needed, but otherwise good to go

raitisbe (approved):

fzadrazil:
Thanks for the review, I will take care of deployment to the test server.

PR #1551, opened 2021-01-29
Get and set layer/feature properties using function and types

raitisbe:
## Layer attribute changes:
Attribution-> attribution
events-suspended -> eventsSuspended
from_composition -> fromComposition
show_in_manager -> showInLayerManager
info_format -> infoFormat
hs-layman-synchronizing -> hsLaymanSynchronizing
metadata -> Now contains urls array to metadata resources and style object.
MaxScaleDenominator -> maxResolutionDenominator
**legendImage** -> legends
**Layer** - Using getCachedCapabilities(..).Layer instead which might contain multiple layers if we are using just a subset of sublayers available on a container layer.

## Removed layer attributes
**BoundingBox**
**Copyright**
**hsFilters**
**img** - (Replaced with already existing 'thumbnail' attribute);
**hsFilters**
**Metadata**
**MetadataURL** - (merged into 'metadata')
**minScale**
**maxScale**
**manuallyAdded**
**position**
**saveState**
## Feature attribute changes:
**unit_id** -> unitId

## Remove feature attributes
**is_hs_composition_extent**

## Needs to be migrated TODO
Feature:
**hstemplate** - (TODO: deprecate or create the custom template functionality since, it's not supported currently);

Layer:
**dimensions_time** - (TODO: remove in favor of dimensions.time which is more in line with WMS-T spec);


projects/hslayers/src/components/layermanager/layermanager-metadata.service.ts
```diff
@@ -63,7 +73,7 @@ export class HsLayerManagerMetadataService {
   }
 
   metadataArray(layer: Layer): Array<any> {
-    const obj = layer.layer.get('MetadataURL');
```
raitisbe, line 66: I'm afraid this wont be so easy. Should explore how the MetadataURL which is a string is used. 'metadata' is an object which can hold anything currently. We could have layer.get('metadata') be an object {key: string, key2: string, url: string} where url was MetadataURL, but have to be reasonably sure we dont break anything.

raitisbe: Turns out MetadataURL is a aray-ish object similar to {'0': {OnlineResource: 'http://..'}}. I Rewrote it to be incorporated into the metadata attribute which is now a more complex object of {urls: [], style: {}}

jmacura:
I see a lot of merge conflicts coming... :-(

> dimensions_time - (TODO: remove in favor of dimensions.time which is more in line with WMS-T spec);

I gonna touch this as a part of the new WMS-t implementation.

raitisbe:
Maybe you can do the actual removal in this or other branch which i can
rebase then? I think the current dimensions_time is used in some old
nonoperational hackathon projects anyways.

jmacura:
Well, until now I intended to preserve backwards-compatibility, but if we know that no operational application is using the old param, then I will just remove it. I have to think for a while, what will be the least painful way, if I sneak some changes here or not...

FilipLeitner (approved):

PR #1552, opened 2021-01-29
Layman transform

jmacura:
Changes to Layman version 1.8 are reflected in this PR.


jmacura:
*Not* ported to 2.x
Ported to 3.x

PR #1559, opened 2021-02-02
Layer cookies optional (3.0)

### LINKED ISSUE #1497: Allow configuration of saving layers to localStorage
* It is sometimes undesired to save&load compositions on map refresh.
* Compositions shall not be saved while closing the window

Relates to 1.27, 2.5, 3.0.

jmacura:
port of #1553

closes #1497

### DIFF
--- a/projects/hslayers/src/components/save-map/save-map.service.ts
+++ b/projects/hslayers/src/components/save-map/save-map.service.ts
@@ -13,7 +13,7 @@ import {
 import {Image as ImageLayer, Tile, Vector as VectorLayer} from 'ol/layer';
 import {Map} from 'ol';
 
-import {HsConfig} from 'hslayers-ng';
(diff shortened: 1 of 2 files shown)


raitisbe:
I think you might have done the same thing @DailisLangovskis already fixed. Could you clarify between yourselves?

DailisLangovskis:
> I think you might have done the same thing @DailisLangovskis already fixed. Could you clarify between yourselves?

I @jmacura implementation is just better, so just remove this.initOnPageReloadListener() line and the function it self and it is all good!

jmacura:
> I think you might have done the same thing @DailisLangovskis already fixed. Could you clarify between yourselves?

You are correct, I did not notice. We have already clarified, thanks for spotting that!

DailisLangovskis (approved):

Issue #1498, opened 2021-01-18
Map gets randomly rotated and mirrored


jmacura:
Steps to reproduce:
1. open https://www.agrihub.cz/mapa
2. zoom to a random place
3. go to https://www.agrihub.cz main page
4. open "Výnosový potenciál" composition from the list below

![image](https://user-images.githubusercontent.com/5598693/104926080-faf58a80-599f-11eb-8dcc-5785e7229f93.png)

relates to version 1.27.2

Reported by @kcharvat in Firefox

jmacura:
I can reproduce it approx. in 1 of 15 attempts.
From the observations so far, it seems that _"something"_ is loaded too soon and the map won't get the correct `center` and `zoom`.

kcharvat:
For me in Firefox, it is very often, yesterday night, it was in every attempt. Now in 1 from 3

raitisbe:
For me the first link doesnt open. Seems like mime type config on the server, because for css it should not be text/html.

![image](https://user-images.githubusercontent.com/395514/105357942-2e355500-5bfe-11eb-98c8-4953678e716e.png)

jmacura:
> 
> 
> For me the first link doesnt open. Seems like mime type config on the server, because for css it should not be text/html.

Seems more like a Liferay bug to me. These content type mismatches with text/html usually happen when the target source does not exist (and the server responses with 404 error).

But this works fine for me (Firefox 84 @ Windows 10)... 😕

kcharvat:
Please is there any progress. This is for us now critical. We have next  week presentation of Innovation Hub and this is key functionality.  For me it is now biggest problem

fzadrazil:
Seems to be fixed by a combination of factors. Only some viewport parameters has been cloned from the default_view that caused maxZoom and other params to be ignored fbbbb91aae7779be2e01d2ab0dd7feb2189a02ca .
After setting extent constrain and multiWorlds=false, the issue doesn't appear anymore or is even harder to reproduce now.

PR #1565, opened 2021-02-03
Fix wrong viewport and layer list (1.27)

fzadrazil:
- Fix broken viewport when loading compositions #1498 
- Fix blank layer manager when added layer causes an error 

Should be also checked for 2.x and 3.x

### DIFF
--- a/components/compositions/compositions-parser.service.js
+++ b/components/compositions/compositions-parser.service.js
@@ -259,7 +259,6 @@ export default function (
       }
       const layers = me.jsonToLayers(obj);
       layers.forEach((lyr) => {
-        $timeout(() => {
           try {
             HsMapService.addLayer(lyr, true);
           } catch (e) {
@@ -267,7 +266,6 @@ export default function (
             $log.warn(error);
             me.createErrorDialog(error);
           }
-        }, 150);
       });
 
       if (angular.isObject(obj.current_base_layer)) {
--- a/components/compositions/compositions-parser.service.js
+++ b/components/compositions/compositions-parser.service.js
@@ -23,7 +23,8 @@ export default function (
   HsCompositionsLayerParserService,
   HsLayoutService,
   $log,
-  $compile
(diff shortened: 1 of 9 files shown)


components/compositions/compositions-parser.service.js
```diff
@@ -258,7 +259,15 @@ export default function (
       }
       const layers = me.jsonToLayers(obj);
       layers.forEach((lyr) => {
-        HsMapService.addLayer(lyr, true);
+        $timeout(() => {
```
raitisbe, line 262: This doesnt seem right. At least the timeout should be outside forEach, se we don't trigger as many digests (at 10 at short interval they fail and break application) as there are layers.

components/layermanager/layermanager.service.js
```diff
@@ -759,9 +780,16 @@ export default function (
   function init() {
     map = HsMapService.map;
     HsMapService.map.getLayers().forEach((lyr) => {
-      layerAdded({
-        element: lyr,
-      });
+      $timeout(() => {
```
raitisbe, line 783: Why is the timeout needed?

raitisbe (changes requested):

components/compositions/compositions-parser.service.js
FilipLeitner: Oh I havent realized that, tested on some 6 layers and it went all well. 

Timeout might not be needed tho...there was a bug on Agrihub in Firefox only once more. Sometimes app.js layer failed to load and it left a whole layermanager empty. Problem was I could replicate much and also taht when putting a breakpoint to the error location it just never bugged. So I tought little  timeout might be enough. but wanted to add try/catch to make sure...but without timeout it would still be almost the same

FilipLeitner:
Before merge we should test on hub just to be sure its all good after the last commit @fzadrazil

raitisbe (approved):

PR #1583, opened 2021-02-08
Issue 1582/missing metadata info

DailisLangovskis:


projects/hslayers/src/components/compositions/endpoints/compositions-status-manager.service.ts
```diff
@@ -50,9 +52,19 @@ export class HsCompositionsStatusManagerService {
       ds.listLoading.unsubscribe();
       delete ds.listLoading;
     }
-    ds.listLoading = this.$http.get(url).subscribe(
+    const listLoading = this.$http.get(url).pipe(
+      timeout(2000),
```
raitisbe, line 56: Why is timeout necessary?

DailisLangovskis: To timeout failed or pending requests, that are slowing down the composition loading.

raitisbe (approved):

PR #1591, opened 2021-02-10
Hide hs logo

DailisLangovskis:
Best I could come up with right now!


FilipLeitner:
I'd say its not that necessary now when sidebar shirks automaticaly. We can keep it just to make sure tho.

PR #1609, opened 2021-02-17
Hsl theme

### LINKED ISSUE #1595: Impressum text shall be also styleable
... like the rest of sidebar. Now it leads to unreadable colour combinations.
image

FilipLeitner:
Runtime bootstrap variable changes based on: https://gist.github.com/johanlef/518a511b2b2f6b96c4f429b3af2f169a#file-_functions-override-scss

fixes #1595


FilipLeitner:
Browser support is determined by https://caniuse.com/calc

projects/hslayers/src/config.service.ts
```diff
@@ -95,7 +96,7 @@ export class HsConfig {
   open_lm_after_comp_loaded?: any;
   draggable_windows?: boolean;
   connectTypes?: any;
-  theme?: any;
+  theme?: string;
```
raitisbe, line 99: Better to have an enum for that

PR #1610, opened 2021-02-18
Release 3.0.1

raitisbe:
Release note staging area: https://docs.google.com/document/d/1M7a6WTnRz80Af1lVrSP2h1ZajWviziVh8OXZZ1XfbK4/edit?usp=sharing


jmacura:
Should be 3.0.0, I suppose?

raitisbe:
> Should be 3.0.0, I suppose?

Does it matter?

jmacura:
> 
> 
> > Should be 3.0.0, I suppose?
> 
> Does it matter?

It makes little sense to skip a number, but otherwise it does not 🙂

raitisbe:
> > > Should be 3.0.0, I suppose?
> > 
> > 
> > Does it matter?
> 
> It makes little sense to skip a number, but otherwise it does not 🙂

You're right probably. The Thing is that I had 3.0.0 for most of the time in summer and autumn, then I published to 3.0.1-rc.0 on npm and now I'm a bit afraid to change it down to 3.0.0, because it might not appear as the last version if we sort it by number instead of publication date.

PR #1597, opened 2021-02-13
Refactor UI for WMS-t layers in LayerManager

jmacura:

PR #1656, opened 2021-03-09
Enhance capability of time-editor

jmacura:
This PR is consecutive to #1597 and re-implements the possibility to use layers with time-points defined with Duration ISO string (like "1999-01-22T19:00:00/2018-01-22T13:00:00/PT8766H") in the new time-editor.

image

Layer on the screenshot is "srazky_l_2000_2018_r" from "https://gis.lesprojekt.cz/cgi-bin/mapserv?map=/home/dima/maps/m_liberec.map".

### DIFF
--- a/projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
+++ b/projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
@@ -300,16 +300,3 @@ export class HsLayerManagerWmstService {
     return timePoints;
   }
 }
-
-declare global {
-  interface Date {
-    monthDiff: (d2) => number;
-  }
-}
-Date.prototype.monthDiff = function (d2) {
-  let months: number;
-  months = (d2.getFullYear() - this.getFullYear()) * 12;
-  months -= this.getMonth() + 1;
-  months += d2.getMonth();
-  return months <= 0 ? 0 : months;
-};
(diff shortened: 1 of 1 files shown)


projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
```diff
@@ -56,8 +57,6 @@ export class HsLayerManagerWmstService {
         dateComponent.search('Y') > -1 ? dateComponent.search('Y') : undefined;
       const indexOfM =
         dateComponent.search('M') > -1 ? dateComponent.search('M') : undefined;
-      const indexOfW =
```
raitisbe, line 59: What was wrong with week?

jmacura: @raitisbe Nothing, it was just unused.

raitisbe: I see. I was just vaguely following the ISO8601 standard: P is the duration designator (for period) placed at the start of the duration representation.
Y is the year designator that follows the value for the number of years.
M is the month designator that follows the value for the number of months.
W is the week designator that follows the value for the number of weeks.
D is the day designator that follows the value for the number of days.

jmacura: No objections here. I just deleted dead code, that's it. Do you request any changes to this PR before merging?

raitisbe: Ok, lets merge

PR #1659, opened 2021-03-12
Styles

FilipLeitner:


raitisbe (approved):

jmacura:
@FilipLeitner @fzadrazil A bit late on the party, but shouldn't the `projects/hslayers-server/src/data/hslayers-server.db` file be gitignored?

fzadrazil:
> @FilipLeitner @fzadrazil A bit late on the party, but shouldn't the `projects/hslayers-server/src/data/hslayers-server.db` file be gitignored?

Can be. I wanted to include it as a template to show how the database looks like. But yes, probably should be ignored.

PR #1686, opened 2021-03-17
Fixes 3x

### LINKED ISSUE #1665: Request datasets from the Micka CSW endpoint.
Only services are requested from Micka at this point. Datasets should be included in the Add data panel. Every dataset can include many urls (can be ZIP, PDF files or even links to web sites). Only first valid WMS should be selected and displayed in Add data.

There is also a new CSW parameter `validservice` which can have values of 0 - not available, 1 - available but not INSPIRE valid, 2 - available and valid. We should query records with `validservice > 0`, eq.
```
https://hub.lesprojekt.cz/micka/csw?query=validservice%3E0&sortby=title%3AA&t=1615363013
```

FilipLeitner:
fixes #1665


projects/hslayers/src/assets/locales/cs.json
```diff
@@ -78,7 +89,8 @@
     },
     "errorWhileRequestingLayers": "Chyba při žádání vrstev ze serveru",
     "capabilitiesParsingProblem": "Problém při načítání informací o službě",
-    "serviceTypeNotMatching": "Typ služby URL se neshoduje s vybranou kategorií"
+    "serviceTypeNotMatching": "Typ služby URL se neshoduje s vybranou kategorií",
+    "Basemap": "Přidat jako podkladovú vrstvu"
```
jmacura, line 93:
```suggestion
    "Basemap": "Přidat jako podkladovou vrstvu"
```

jmacura (commented):
I would think that our currently preferred approach is **not** to add translations to foreign languages (LV in this case) blindly, but let the i18n Ally extension warn the mates about missing translations to do.

FilipLeitner:
> Besides the typo:
> I would think that our currently preferred approach is **not** to add translations to foreign languages (LV in this case) blindly, but let the i18n Ally extension warn the mates about missing translations to do.

Ok, wasnt aware of such a warnings so I tought itll be safer to add it in english at least

projects/hslayers/src/components/layermanager/layermanager.service.ts
```diff
@@ -925,13 +947,54 @@ export class HsLayerManagerService {
...
+
+    if (pathLayers.length > 0) {
+      //Get max avaialble index value
+      const maxPathZIndex = Math.max(
+        ...pathLayers.map((lyr) => lyr.layer.getZIndex() || 0)
+      );
+
+      layer.setZIndex(maxPathZIndex + 1);
+      //Increase zIndex of the layer that are supposed to be rendered above inserted
+      for (const lyr of this.data.layers.filter(
```
raitisbe, line 972: Could have used HsLayermanagerPhysicalListService.moveTo

FilipLeitner: Id also have to move the function out of the HsLayermanagerPhysicalListService becasue of the circular dependency

raitisbe: right

projects/hslayers/src/components/save-map/save-map.service.ts
```diff
@@ -172,30 +167,26 @@ export class HsSaveMapService {
    * @function layer2json
    * @param {Array} layers All map layers
    * @param {Array} tickedLayers List of layers and if they have been ticked
-   * @returns {Array} JSON object representing the layers
+   * @return {Array} JSON object representing the layers
    */
-  layers2json(layers, tickedLayers?) {
+  layers2json(layers, onlyTicked?: boolean) {
```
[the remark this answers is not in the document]
FilipLeitner: Ye I can see it now. Itll be definately easier to jsut pass already filtered array. Ill add a quick fix for this. Thanks for pointing it out

raitisbe (approved):

PR #1700, opened 2021-03-26
Basic drag n drop functionality for physical layer list

FilipLeitner:


projects/hslayers/src/components/draw/draw.service.ts
```diff
@@ -186,7 +186,6 @@ export class HsDrawService {
         format: layman ? 'hs.format.WFS' : null,
         url: layman ? layman.url + '/wfs' : null,
       },
-      zIndex: Infinity,
```
raitisbe, line 189: Noticed this too and fixed in hslayers-ng-11 branch. Instead of dropping the zIndex I calculated the current max zindex and added 1. Maybe you can downport the changes to this branch to not get conflicts later?

raitisbe (approved):

PR #1701, opened 2021-03-28
Fix ngStyle of layer legend (2.5)

jmacura:
This hotfix is against the obsolete version 2, but I was just working on that version and I felt it makes no harm to fix that if eventually there would come another point release in v2.
~~I suspect this fix would need a port to v3, but have to check first.~~


raitisbe:
Thanks this looks much nicer. Was it working before, just not looking good, or it was broken?

raitisbe (approved):

jmacura:
Broken actually.
Btw, not sure what was this code's original purpose. I used it to sneak in a custom legend created programatically during run time.

PR #1702, opened 2021-03-29
Legend layer fixes

jmacura:
port of #1701

### DIFF
--- a/projects/hslayers/src/components/layermanager/layermanager.service.ts
+++ b/projects/hslayers/src/components/layermanager/layermanager.service.ts
@@ -11,6 +11,7 @@ import {ImageWMS, TileArcGISRest, TileWMS} from 'ol/source';
 import {Injectable} from '@angular/core';
 import {METERS_PER_UNIT} from 'ol/proj';
 
+import {HsBaseLayerDescriptor} from './base-layer-descriptor.interface';
 import {HsConfig} from '../../config.service';
 import {HsDrawService} from '../draw/draw.service';
 import {HsEventBusService} from '../core/event-bus.service';
@@ -57,7 +58,7 @@ export class HsLayerManagerService {
   data: {
     folders: any;
     layers: HsLayerDescriptor[];
-    baselayers: any[];
+    baselayers: HsBaseLayerDescriptor[];
     terrainlayers: any[];
     baselayersVisible: boolean;
     baselayer?: string;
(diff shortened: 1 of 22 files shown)


jmacura:
TODO: sneak in a fix for print
`[(ngModel)]="title" [ngModelOptions]="{standalone: true}"`

jmacura:
This PR adds new and quite generic pipe called FilterPipe. It can be used by implementing a callback function of pattern Object => boolean

fixes #1717

projects/hslayers/src/components/layermanager/layermanager.service.ts
```diff
@@ -323,6 +327,11 @@ export class HsLayerManagerService {
     });
   }
 
+  refreshLists(): void {
```
raitisbe, line 316: Can you add some comment why was cloning necessary, because I think we will forget otherwise?

projects/hslayers/src/components/print/print.service.ts
```diff
@@ -15,8 +15,18 @@ export class HsPrintService {
...
    */
   print(title: string): void {
-    const canvas = this.HsMapService.getCanvas();
-    const img = canvas.toDataURL('image/png');
+    const canvases = this.HsMapService.getCanvases();
+    const composition = document.createElement('canvas');
+    composition.width = canvases[0].clientWidth;
+    composition.height = canvases[0].clientHeight;
+
+    canvases.forEach((canvas) => {
```
raitisbe, line 23: Realy nice!

projects/hslayers/src/components/legend/legend.component.ts
```diff
@@ -34,6 +34,7 @@ export class HsLegendComponent {
     const descriptor = this.HsLegendService.getLayerLegendDescriptor(layer);
     if (descriptor) {
       this.layerDescriptors.push(descriptor);
+      this.refreshList();
```
raitisbe, line 35: Should we repaint the whole list if just one element is added?

DailisLangovskis (approved):

projects/hslayers/src/components/print/print.service.ts
jmacura: @FilipLeitner 's work 😉

projects/hslayers/src/components/layermanager/layermanager.service.ts
jmacura: Good point, gonna do that.

jmacura: Reflected in 26ee014

PR #1716, opened 2021-03-31
Fix forgotten AngularJS syntax (2.5)

jmacura:
Similar tiny fix like #1701


raitisbe (approved):

jmacura:
@raitisbe Oh, I was investigating yet another issue of this kind I want to add to this PR, you are too fast to my pace 😄 But it's my bad, I should have kept it as a draft until ready.

PR #1733, opened 2021-04-09
fix(hslayers-app): Fix hslayers-app build

FilipLeitner:
Target es2015 in hslayers-app to prevent Error: getInternalNameOfClass() called on a non-ES5 class:
expected CookieService to have an inner class declaration

was getting the error on both hub.lesprojekt and local

image


raitisbe:
No need to target es2015 as per https://stackoverflow.com/questions/62733064/error-in-getinternalnameofclass-called-on-a-non-es5-class-expected-angularfi second answer.

But several things needed to be done on server:
1) Don't use sudo to `npm ci`, but have a linux group own the hslayers directory with write rights
2) run `node_modules/.bin/ngcc` after recreating node_modules with `npm ci`

Also needed to create a blank projects/hslayers-app/src/locale directory which should not be necessary in future with the removal of moment dependency.

PR #1745, opened 2021-04-13
Issue 1741 snap

FilipLeitner:


projects/hslayers/src/assets/locales/en.json
```diff
@@ -376,7 +376,9 @@
     "saveDrawingToLayer": "Save drawing to layer",
     "Select layer": "Select layer",
     "serverLayers": "Server layers",
-    "unsavedDrawing": "Unsaved drawing"
+    "unsavedDrawing": "Unsaved drawing",
+    "allowSnapping": "Allow snapping",
+    "snapTo": "Snap to features of layer:"
```
jmacura, line 381: I would leave the colon out of the translation string. For translating convenience.

projects/hslayers/src/components/draw/draw.service.ts
```diff
@@ -52,6 +52,13 @@ export class HsDrawService {
   hasSomeDrawables: boolean;
   draw: Draw;
   modify: any;
+
+  //Snap interaction
+  snap: Snap;
+  snapActive = false;
+  snapSource: any;
```
raitisbe, line 59: VectorLayer

projects/hslayers/src/assets/locales/en.json
FilipLeitner: e7301c4

projects/hslayers/src/components/draw/draw.service.ts
FilipLeitner: d9db600

FilipLeitner:
Should also note that for now its only available in draw panel. I feel like toolbar is already filled up with various functionality  and  adding another would make it bit opaque. So I was a kinda reluctant to add it there. What do you think @raitisbe  ?

projects/hslayers/src/components/draw/draw.service.ts
```diff
@@ -612,6 +640,50 @@ export class HsDrawService {
...
+
+  /**
+   * Handle snap interaction changes
+   * Remove snap interaction if it already exists, recreate it if source is provided.
+   */
+  toggleSnapping(source?: VectorSource): void {
+    this.HsMapService.loaded().then((map) => {
+      this.snapSource = source ? source : this.snapSource;
+      if (this.snapActive && this.snapSource) {
+        if (this.snap) {
```
raitisbe, line 660: There seems to be unnecessary duplication

projects/hslayers/src/components/draw/draw.service.ts
```diff
@@ -612,6 +640,50 @@ export class HsDrawService {
...
+        }
+      }
+    });
+  }
+  /**
+   * Changes layer source of snap interaction
+   */
+  changeSnapSource(layer: Layer): void {
+    const isLayerClustered = this.HsLayerUtilsService.isLayerClustered(layer);
+    let snapSourceToBeUsed;
```
raitisbe, line 680: Ternary operator?

FilipLeitner: Sure, itll be cleaner. I just coppied this part from else where so I might as well refactor both occurences.

raitisbe: great, thanks

raitisbe (approved):

PR #1753, opened 2021-04-19
build: Upgrade to Angular 10

raitisbe:
To upgrade from ng 9.x use `ng update @angular/core@10 @angular/cli@10`

BREAKING CHANGE: See angular upgrade guide: https://update.angular.io/?l=2&v=9.1-10.2

re #1752


fzadrazil (approved):
I tried upgrading full and simple apps and it seems everything is working as before.

PR #1756, opened 2021-04-20
feat: Create time dimension def from getCaps

raitisbe:
Populate layers time dimension definition if
it has none, but GetCapabilities has it.

### DIFF
--- a/projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
+++ b/projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
@@ -109,7 +109,7 @@ export class HsLayerManagerWmstService {
    * @returns True for WMS layer with time support
    */
   layerIsWmsT(layer: HsLayerDescriptor | Layer): boolean {
-    const olLayer: Layer = layer?.layer ? layer.layer : layer;
+    const olLayer: Layer = layer.layer ?? layer;
     if (!olLayer) {
       return false;
     }
--- a/projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
+++ b/projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
@@ -1,4 +1,3 @@
-import moment from 'moment';
 import {Injectable} from '@angular/core';
 
 import {Layer} from 'ol/layer';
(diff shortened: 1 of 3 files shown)


jmacura (commented):
If I got correctly the outcome of these changes, you are actually trying to achieve a very opposite business logic than me.
My concern was not to pollute OL's `Layer` object with so many custom properties and keep all necessary data in our wrappers: `HsLayerDescriptor` and `HsDimensionDescriptor`. The config values provided in OL's `Layer` are read as early as possible and no longer used.

So I wonder, why are these changes necessary?

raitisbe:
> If I got correctly the outcome of these changes, you are actually trying to achieve a very opposite business logic than me.
> My concern was not to pollute OL's `Layer` object with so many custom properties and keep all necessary data in our wrappers: `HsLayerDescriptor` and `HsDimensionDescriptor`. The config values provided in OL's `Layer` are read as early as possible and no longer used.
> 
> So I wonder, why are these changes necessary?

The dimensions objects had to be set because we are checking for it in layermanager-wmst.service.ts  
 setLayerTime(currentLayer: HsLayerDescriptor, newTime: string): void {
    ...
    const dimensions = getDimensions(currentLayer.layer);
    const dimensionDesc = new HsDimensionDescriptor(
      'time',
      dimensions['time'] || dimensions['TIME']
    );

Otherwise it was failing with dimensions variable undefined error.

Regarding the fillAvailableTimes in OnInit It was behaving strange when I had 6 layers with time dimension. For first 4 it was working fine, but last 2 which got its dimensions OL attribute populated later, were displaying Time outside available range warning in the dropdown, because `if (this.availableTimes === undefined && this.layer.uid === layer.uid) {` never got evaluated as true.

Also constructor in layermanager-time-editor.component.ts and `this.hsEventBusService.layerTimeChanges.subscribe ` was executed after the getcapabilities were received and dimension attribute created.

raitisbe:
I just spotted that I'm calling this.HsEventBusService.layerTimeChanges.next twice unnecessarily.

raitisbe:
> My concern was not to pollute OL's `Layer` object with so many custom properties and keep all necessary data in our 
That sounds good, but this attribute is what we are currently using to display the time-editor component with layerIsWmsT() function or am I missing something?

jmacura:
> 
> 
> > My concern was not to pollute OL's `Layer` object with so many custom properties and keep all necessary data in our
> 
> That sounds good, but this attribute is what we are currently using to display the time-editor component with layerIsWmsT() function or am I missing something?

The logic of layerIsWmsT() is to return true only if the time capability is turned on by the config in OL `Layer` def. The content of that config is although arbitrary as 
```
if (getDimensions(layer)?.time) {
      return true;
    }
```

If I remember correctly, we had some discussion with @fzadrazil about this some time ago and the decision was to **not** display time-editor when the WMS has that capability but user won't allow it via that part of config in `Layer`. I might be wrong or we can reconsider this decision as well.

raitisbe:
Changed my approach to use HsLayersDescriptor.time instead of getDimensions/setDimensions. Now its a bit cleaner too.

fzadrazil:
> If I remember correctly, we had some discussion with @fzadrazil about this some time ago and the decision was to **not** display time-editor when the WMS has that capability but user won't allow it via that part of config in `Layer`. I might be wrong or we can reconsider this decision as well.

Well, it is probably not necessary to be able to turn off time editor in config as I think about it now. 

> The logic of layerIsWmsT() is to return true only if the time capability is turned on by the config in OL Layer def.

And what about layers added via Add data panel? You automatically add time editor for those, right?

raitisbe:
> > If I remember correctly, we had some discussion with @fzadrazil about this some time ago and the decision was to **not** display time-editor when the WMS has that capability but user won't allow it via that part of config in `Layer`. I might be wrong or we can reconsider this decision as well.
> 
> Well, it is probably not necessary to be able to turn off time editor in config as I think about it now, if it helps the case.
> 
> > The logic of layerIsWmsT() is to return true only if the time capability is turned on by the config in OL Layer def.
> 
> And what about layers added via Add data panel? You automatically add time editor for those, right?

From Add data panel yes, dimensions attribute for the OL layer is added like it was before, but even if I disabled it, it should still be populated from the GetCapabilities after some delay whith the background process. Im just not sure what would happen with the other non-time related dimensions, probably they would not work.

projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
```diff
@@ -110,39 +109,39 @@ export class HsLayerManagerWmstService {
    * @returns True for WMS layer with time support
    */
   layerIsWmsT(layer: HsLayerDescriptor | Layer): boolean {
-    if (layer?.layer) {
-      // case of HsLayerDescriptor, Layer would have 'sublayers' property instead
-      layer = layer.layer;
-    }
-    if (!layer) {
+    const olLayer: Layer = layer?.layer ? layer.layer : layer;
```
jmacura, line 112: Could have been

```suggestion
    const olLayer: Layer = layer.layer ?? layer;
```

But I am fine with your change as well 🙂

jmacura (approved):
Ok, one comment, but good to go. Thanks.

projects/hslayers/src/components/layermanager/layermanager-wmst.service.ts
raitisbe: Nice trick!
https://youtu.be/UmmOJx_Hxto?t=126

PR #1757, opened 2021-04-20
UI fixes + Micka redirect for full composition metadata

FilipLeitner:


projects/hslayers/src/components/sidebar/sidebar.service.ts
```diff
@@ -256,8 +256,13 @@ export class HsSidebarService {
       }, 550);
     });
 
-    this.HsEventBusService.layoutResizes.subscribe(() => {
+    this.HsEventBusService.layoutLoads.subscribe(() => {
       this.setButtonVisibility();
+
+      //After initial run update sidebar with each layoutResizes event
+      this.HsEventBusService.layoutResizes.subscribe(() => {
+        this.setButtonVisibility();
```
raitisbe, line 264: Will this not stop adjusting button state for different window sizes? Or we now have another mechanism to deal with that?

FilipLeitner: No its basicaly the same as previously just moved the first run of 'setButtonVisibility' from component constructor to  layoutLoads event. 
And to make sure we are not unnecessarily triggering it before that it is subscribed to layoutResizes from layoutLoads event also.

raitisbe (approved):

PR #1770, opened 2021-04-26
ci: Bump NodeJS to 12.x

jmacura:
Node 10 goes end-of-life. Node 12 is the oldest maintained version from now on.

https://nodejs.org/en/about/releases/


raitisbe:
Definitely needed to upgrade, due to better RAM reserving on build startup too.

raitisbe (approved):

PR #1796, opened 2021-05-06
feature: Locales in languages panel can be configured

DailisLangovskis:
HsConfig enabledLanguages can be set as string from any container app. User must input required languages code name, such as, LV for latvian. Each code must be sepperated using comma.
 Example:
enabledLanguages: 'lv, cs, sk',

### DIFF
--- a/projects/hslayers/src/components/core/core.service.ts
+++ b/projects/hslayers/src/components/core/core.service.ts
@@ -43,8 +43,10 @@ export class HsCoreService {
     } else {
       this.HsLayoutService.sidebarExpanded = true;
     }
-
-    this.translate.addLangs(['en', 'cs', 'lv', 'sk']);
+    const languages = this.HsConfig.enabledLanguages
+      ? this.HsConfig.enabledLanguages.split(',').map((lang) => lang.trim())
+      : ['cs', 'lv'];
+    this.translate.addLangs(languages);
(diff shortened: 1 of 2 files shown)


projects/hslayers/src/components/core/core.service.ts
```diff
@@ -43,8 +43,10 @@ export class HsCoreService {
     } else {
       this.HsLayoutService.sidebarExpanded = true;
     }
-
-    this.translate.addLangs(['en', 'cs', 'lv', 'sk']);
+    const languages = this.HsConfig.enabledLanguages
+      ? this.HsConfig.enabledLanguages.split(',').map((lang) => lang.trim())
+      : ['cs', 'lv'];
```
jmacura, line 48: Are you disabling 'en' in this line?

DailisLangovskis:
No, en is enabled always by default

jmacura ***@***.***> (šajā datumā: ceturtd., 2021. g. 6.
maijs 14:30) rakstīja:

projects/hslayers/src/components/core/core.service.ts
jmacura: Thanks for clarification, I suspected you know what you are doing 🙂

FilipLeitner (approved):

Issue #1842, opened 2021-05-14
Layer permissions do not propagate to Layman


fzadrazil:
Setting Public option for newly added layer (eg. SHP) does not store EVERYONE to Layman layer permissions.

PR #1848, opened 2021-05-17
fix: Send correct access rights parameters to layman

FilipLeitner:
Make sure user can subimt only logicaly valid combination of access rights and unify this behavior acros the library

fixes:#1842

### DIFF
--- a/projects/hslayers/src/common/layman/access-rights/layman-access-rights.component.ts
+++ b/projects/hslayers/src/common/layman/access-rights/layman-access-rights.component.ts
@@ -1,0 +1,26 @@
+import {Component, Input, ViewRef} from '@angular/core';
+import {accessRightsInterface} from '../../../components/add-data/common/access-rights.interface';
+
+@Component({
+  selector: 'hs-layman-access-rights',
+  templateUrl: './layman-access-rights.html',
+})
+export class HsCommonLaymanAccessRightsComponent {
+  @Input() access_rights: accessRightsInterface;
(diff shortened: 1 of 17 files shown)


raitisbe:
I see quite a lot of duplication. Maybe this is a good chance to refactor a bit?

projects/hslayers/src/common/layman/access-rights/layman-access-rights.component.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Component({
+  selector: 'hs-layman-access-rights',
+  templateUrl: './layman-access-rights.html',
+})
+export class HsCommonLaymanAccessRightsComponent {
+  @Input() access_rights: accessRightsInterface;
+  constructor() {}
+
+  accessRightChanged(
```
raitisbe, line 12: We could export this as a function nd then reuse it in other components (draw-layer-metadata.component.ts , add-data-vector.component.ts) (in save-map-advanced-form.component.ts looks similar too, but you know better)

FilipLeitner: Ye sure I forgot about save-map-advanced-form.component.ts. Should be used there as well.  Also forgot to delete  accessRightChanged functions from components other than newly created one... thanks for pointing this out. Was kinda busy and trying to rush everything so I do these unnecessary mistakes

raitisbe (approved):

PR #1877, opened 2021-05-25
Issue 1822/styler rewrite

raitisbe:


DailisLangovskis:
err5
err1
err2
These are bugs I encountered during brief testing on map whiteboard app.

DailisLangovskis:
@raitisbe Also is this how dash pattern adding should be like ?
err4
Lastly, it would be nice if the dropdown would stay on the left side of the panel for this selection.
err3

raitisbe:
> @raitisbe Also is this how dash pattern adding should be like ?

Pretty much. Its a list of numbers, just have to make the textboxes smaller and introduce some margins. Usually you dont want more than 2-3 numbers here.

raitisbe:
> err5
> err1
> err2
> These are bugs I encountered during brief testing on map whiteboard app.

Second image error should be fixed. Maybe also first, dont know how it happened. I fixed similar thing when toggling on/off "cluster"  checkbox in layermanager.

raitisbe:
> Lastly, it would be nice if the dropdown would stay on the left side of the panel for this selection.

Interesting. Is dropup not working as needed?
For me it looked like this
image

DailisLangovskis:
> > err2
> > These are bugs I encountered during brief testing on map whiteboard app.
> 
> Second image error should be fixed. Maybe also first, dont know how it happened. I fixed similar thing when toggling on/off "cluster" checkbox in layermanager.

This issue sometimes still pops up, when layer with features and no style property is added to map composition in whiteboard app. Also there was a weird issue, that line symbolizer would not work saying cannot call split function from undefined, but I cannot recreate it again unfortunately.

DailisLangovskis:
> > err5
> > These are bugs I encountered during brief testing on map whiteboard app.
> 
> Second image error should be fixed. Maybe also first, dont know how it happened. I fixed similar thing when toggling on/off "cluster" checkbox in layermanager.

Still getting this object object as cluster text rather then having a number. It can be recreated by adding a new layer, then adding features and pressing cluster the layer checkbox.

DailisLangovskis:
@raitisbe Found another issue regarding declutter. When setting declutter to true, the layer gets deleted from map composition.

raitisbe:
> @raitisbe Found another issue regarding declutter. When setting declutter to true, the layer gets deleted from the map, but is doesn't call the removeLayer listener so it stays saved on the mongo database. However, the features gets removed from database.

This seems to be an issue seperate from styling

DailisLangovskis:
@raitisbe Adding only a line symbolizer rule, somehow forbids users to add point features on the map.

raitisbe:
They are probably just not visible since no symbolizer exists - mark or ico
symbolizer i mean. Should issue a warning toast in that case.

DailisLangovskis (approved):

PR #1881, opened 2021-05-26
Issue 900/wms crs filip

FilipLeitner:
Its working kinda...had to remove one of the wms URLs  cause it was taking server too long to respond so the test failed on timeout. (even though I doubled the default value).


jmacura:
Wow, great job, thanks a lot!
I specifically picked the stanford.edu endpoint, because it contains plenty of records, so it seemed like a proper tough test 😄

jmacura (approved):

FilipLeitner:
> Wow, great job, thanks a lot!
> I specifically picked the stanford.edu endpoint, because it contains plenty of records, so it seemed like a proper tough test 😄

Well in that case we can increase jasmine.DEFAULT_TIMEOUT_INTERVAL even further. But unless there are some other specificities I thkin its not necessary

PR #1887, opened 2021-05-26
Some of add data fixes

DailisLangovskis:


projects/hslayers/src/components/core/event-bus.service.ts
```diff
@@ -162,5 +162,9 @@ export class HsEventBusService {
   layerSelectedFromUrl: BehaviorSubject<VectorLayer> = new BehaviorSubject(
     null
   );
+  /**
+   * Cancels any external url data request from datasources panel
+   */
+  cancelUrlRequest: Subject<void> = new Subject<void>();
```
raitisbe, line 168: Maybe better would be to put it in add-data.service ? To group these subjects by topic.

raitisbe (changes requested):

DailisLangovskis:
You might be right, just wanted to use service, that all components are
already using without having to inject anything else.

Raitis Berzins ***@***.***> (šajā datumā: piektd., 2021. g.
28. maijs 00:29) rakstīja:

projects/hslayers/src/components/core/event-bus.service.ts
DailisLangovskis: Implemented the requested change.

raitisbe (approved):

PR #1903, opened 2021-06-02
Allow to disable automatic legend

jmacura:
Useful when automatic legend does not represent the map naturally, like in the rural-attractiveness apps.
I wanted to implement this before (in #1702) but forgot to finish it. Hence this PR is more than simple.


raitisbe:
Maybe use HsLayerExt functions just to be consistent?

jmacura:
Good point.

raitisbe (approved):

PR #1910, opened 2021-06-04
fix: Toast message width

### LINKED ISSUE #1909: Toast messages stretch based on the length of the text
It even overflows the screen when the text is long enough. Should have fixed size. It seems to appear only in full map, not simple.

FilipLeitner:
closes #1909


raitisbe:
Was the default .hsl .toast {min-width which comes with bootstrap overriden by liferay?

FilipLeitner:
> Was the default .hsl .toast {min-width which comes with bootstrap overriden by liferay?

Well it somehow wasnt even applied.

raitisbe (approved):

PR #1911, opened 2021-06-04
fix(routing): Don't add waypoints on draw and measure

raitisbe:


raitisbe:
> between

ok

jmacura (approved):
As far as this PR goes, I think it's great!
If you'd like to incorporate fix for #1925 in this PR, or leave it for another, that's up to you.

PR #1918, opened 2021-06-07
feat(material): extend LayerManagerComponent

raitisbe:
This is just an experiment demonstrating extension of Hslayers components with different template and class members


SLeitgeb (approved):

raitisbe:
Exactly this. So we dont reinvent the wheel and also make synchronizing
develeopment of material and bootstrap component versions easier

PR #1939, opened 2021-06-18
Fix sublayers toggle

### LINKED ISSUE #1873: Toggle sublayers button does not toggle
image

These layers are from https://wms.geo.bs.ch?

jmacura:
A very streamline fix. There was no other class bound when the `expanded` property was set to `false`, so this change simplifies it a bit.
Other places where the 'show' class occurs are worth the check.

fixes #1873


raitisbe (approved):

jmacura:
> Other places where the 'show' class occurs are worth the check.

Checked and are OK.

PR #1943, opened 2021-06-21
fix: Unreachable side-buttons on small screens

### LINKED ISSUE #1940: Unreachable side-buttons on small screens
On a small screen:
image
Note, that there is nowhere to scroll.

The same app on a larger screen:
image

Tested with 'full' app on Firefox 89.

FilipLeitner:
fixes #1940


jmacura (approved):
Awesome, thanks!

Issue #1502, opened 2021-01-19
Expose layer type and URL (if applicable) in LM settings


jmacura:
Here somewhere
![image](https://user-images.githubusercontent.com/5598693/104972758-fe663180-59f2-11eb-8d1c-8701006d6c23.png)

would be much easier to debug layer-source issues then...

The information can have e.g. a form like:
_WMS layer from https://blabla.com/sth_
_GeoJSON layer from http://example.com/layer.json_

Other UI suggestions welcomed.

PR #1947, opened 2021-06-21
Expose layer source type and URL in LayerManager

jmacura:
image
---
image
---
image
---
image


jmacura:
I kindly ask for translations in LV and SK, @DailisLangovskis @FilipLeitner 🙂

jmacura:
resolves #1502

projects/hslayers/src/components/utils/layer-utils.service.ts
```diff
@@ -191,12 +193,58 @@ export class HsLayerUtilsService {
   }
 
   /**
-   * @ngdoc method
-   * @name HsLayerUtilsService#isLayerInManager
-   * @param {Ol.layer} layer Layer to check
-   * @returns {boolean} True if showInLayerManager attribute is set to true
-   * @description Test if layer is shown in layer switcher
+   * Test if the features in the vector layer come from a GeoJSON source
```
DailisLangovskis, line 196: I would might try testing if layer is vector layer with utils function and then use switch case to get the format to make code shorter, but I don't know if there is a special need for this format.

raitisbe:
Good job guys! Can I ask @jmacura to squash the style commits into one, to not make the commit log unnecessary long?

jmacura:
> 
> 
> Good job guys! Can I ask @jmacura to squash the style commits into one, to not make the commit log unnecessary long?

Thanks. I will try my best.

jmacura:
> 
> 
> > Good job guys! Can I ask @jmacura to squash the style commits into one, to not make the commit log unnecessary long?
> 
> Thanks. I will try my best.

@raitisbe I have reduced the number of commits to half. Wanted to preserve the original commit author so I kept Filip's and Dailis' commits separate (I don't think it's possible in git to record more then 2 commit authors).

projects/hslayers/src/components/utils/layer-utils.service.ts
jmacura: Good point. I generally wanted to keep the same logic of boolean-ish functions. That was the only reason really. Your proposal would be more concise of course.

jmacura: I have refactored. It is not exactly as you meant (if I am correct), but it's at least shorter.

jmacura:
Good to go from my site.

raitisbe (approved):

PR #1953, opened 2021-06-27
Potential fix - do not  add null or undefined url to history list

DailisLangovskis:
@jmacura Would appreciate if You could try to test if the values null or undefined still pops up in the history list dropdown.
This fix does not guaranty that it will work, but at least could be a starting point.


jmacura (approved):
Provided we do not know the steps to reproduce this bug, it is kind of hard to tell, if the fix is fixing it 😄 
But I have tried adding multiple WMS/WFS/GeoJSON sources via URL as well as some data from catalogues and no `null` or `undefined` appeared. So far I can tell it does not make the things worse and it might be the fix.

PR #1965, opened 2021-07-01
Layman token refresh fix

fzadrazil:
* simplify token refresh workflow
* move duplicit error notification to common service
* automatically logout from Layman when token become invalid
* change Login/Logout button style


projects/hslayers/package.json
```diff
@@ -56,6 +56,9 @@
     "xml-js": "^1.6.11"
   },
   "dependencies": {
-    "tslib": "^2.0.0"
+    "@angular/platform-browser": "^12.1.0",
```
DailisLangovskis, line 59: @fzadrazil I suppose these deps were not intended.

fzadrazil: Sure not, I have also noticed and will push a fix in a minute.

jmacura:
👏 for moving the duplicate function into common/.

DailisLangovskis (approved):

FilipLeitner (approved):

PR #1972, opened 2021-07-06
Filip issue 1954

FilipLeitner:


projects/hslayers/src/components/draw/draw-layer-metadata.component.ts
```diff
@@ -126,5 +125,8 @@ export class HsDrawLayerMetadataDialogComponent
 
   selectLayer(layer): void {
     this.data.selectLayer(layer);
+    setTimeout(() => {
```
DailisLangovskis, line 128: Why do we need this timeout and dialog closure after selecting a layer?

FilipLeitner: We dont, it just felt better as it takes some time to get layer trom Layman

DailisLangovskis: It would be better to remove it, because the layer will get loaded anyways. What the timeout does, it gives enough time for user to click the layer second time and two identical layers gets added to the map.

FilipLeitner: Sure sounds reasonable , thanks :)

PR #1991, opened 2021-07-13
Issue 840

DailisLangovskis:
Need assist on remaining failing tests.
@jmacura  ->
 // this.HsLog.error(
    //   `Cannot decide a type of source of layer ${getTitle(layer)}`
    // ); 
This is the last remaining cause of test fails, because for some reason loading composition from test/data file, doesn't successfully find the layer type.

### DIFF
--- a/projects/hslayers/src/testing-utils.ts
+++ b/projects/hslayers/src/testing-utils.ts
@@ -1,0 +1,13 @@
+const orgConsoleError = window.console.error;
+
+export function patchConsoleToFailOnError(): void {
+  window.console.error = function (...args: any[]) {
+    orgConsoleError.apply(this, args);
+    try {
+      throw new Error('console.error');
+    } catch (err) {
+      window.console.info('console.error', args, err);
+    }
+    fail('console.error was called, this is not allowed in a unit test run');
+  };
+}
--- a/projects/hslayers/karma.conf.js
+++ b/projects/hslayers/karma.conf.js
@@ -13,7 +13,10 @@ module.exports = function (config) {
       require('@angular-devkit/build-angular/plugins/karma')
     ],
     client: {
-      clearContext: false // leave Jasmine Spec Runner output visible in browser
+      clearContext: false, // leave Jasmine Spec Runner output visible in browser
+      jasmine:{
+        random: false
+      }
     },
     coverageIstanbulReporter: {
       dir: require('path').join(__dirname, '../../coverage/hslayers'),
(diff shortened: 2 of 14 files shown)


jmacura:
Shall read
```
this.HsLog.warn(
    Cannot decide a type of source of layer ${getTitle(layer)}
);
```

It is not an error, it should only produce warning. My mistake, apologies!

FilipLeitner:
> Shall read
> 
> ```
> this.HsLog.warn(
>     Cannot decide a type of source of layer ${getTitle(layer)}
> );
> ```
> 
> It is not an error, it should only produce warning. My mistake, apologies!

With error it fails beacuse of this function:
https://github.com/hslayers/hslayers-ng/blob/develop/projects/hslayers/src/components/utils/layer-utils.service.mock.ts#L7

We might want to mock it better

DailisLangovskis:
@raitisbe The last fails happen from -> this.styleObject = await this.sldToJson(sld); on the third-party lib 
_stack: "TypeError: Cannot read property 'forEach' of undefined\n    at http://localhost:9876/_karma_webpack_/webpack:/projects/hslayers/node_modules/geostyler-sld-parser/build/dist/SldStyleParser.js:1090:1\n_
I am not sure, how to fix this, would appreciate some help.

FilipLeitner:
> @raitisbe The last failing tests arises from -> this.styleObject = await this.sldToJson(sld); on the third-party lib
> _stack: "TypeError: Cannot read property 'forEach' of undefined\n at [http://localhost:9876/_karma_webpack_/webpack:/projects/hslayers/node_modules/geostyler-sld-parser/build/dist/SldStyleParser.js:1090:1\n](http://localhost:9876/_karma_webpack_/webpack:/projects/hslayers/node_modules/geostyler-sld-parser/build/dist/SldStyleParser.js:1090:1%5Cn)_
> I am not sure, how to fix this, would appreciate some help.

Tracked the error down to be cause by this [function ](https://github.com/hslayers/hslayers-ng/blob/develop/projects/hslayers/src/components/styles/styler.spec.ts#L104)call ..in some specific occasions. or rather when styler tests were run in some specific sequence. Havent got exacty to that 'forEach' call but when test were not run in order it failed and Vector layerd had sld already defined (seemed like sld from 'SLD should be generated from OL style' test ...so it might be that test runnig before caused that)
...so it was basically either to define vector layer before each test or run tests in order (which I tought might be better ...in general)

raitisbe:
Or could clear/reset the layer data in the failing test

projects/hslayers/src/testing-utils.ts
```diff
@@ -0,0 +1,13 @@
+const orgConsoleError = window.console.error;
+
+export function patchConsoleToFailOnError(): void {
```
raitisbe, line 3: Nice one

raitisbe (approved):

Issue #1944, opened 2021-06-21
originalDimension is undefined


jmacura:
![image](https://user-images.githubusercontent.com/5598693/122806926-3ee01900-d2cb-11eb-9d08-d5b317aec741.png)

PR #1997, opened 2021-07-13
Dimensions refactored

### LINKED ISSUE #1868: WMS-t bug
image

https://wms.geo.bs.ch?

keeping it here now just for future reference

jmacura:
### Before:
image
image
### After:
image

resolves #1868

TODO: ~#1944~(?) ~#2033~ ~#1995~


jmacura:
### Before:
image

### After:
image

resolves #1995

jmacura:
Now rebased on top of #2082

raitisbe:
image
I have added  a layer to test-app, which has some default time values, that get updated later on. The change is broadcasted using layerDimensionDefinitionChanges subject, but instead of the new values appearing the dropdown control gets emptied completely.
It's also a bit hard to create HsLayerWrapper from container application, because its not exported in public-api.
Now using `{layer: {layer}}` when calling next.

raitisbe:
Got some text overlaying the map when the sidebar is on the left side 
image

jmacura:
> It's also a bit hard to create HsLayerWrapper from container application, because its not exported in public-api.
> Now using `{layer: {layer}}` when calling next.

This is an essential design problem. Do we want to prefer HsLayerDescriptor or Layer? Both has advantages and disadvantages.
In layermanager-time-editor I have preferred the HsLayerDescriptor, so I could rely on a strongly typed parametres and don't have to re-check the values every time they are read, whether some other service has not changed the values back into some dirty string or whatever. Other reason was not to pollute the Layer object with more custom properties than necessary (and the dimensions property is a bit clumsy to work with, as it is a plain K-V object, not an array).
And the decision has shortcomings, of course. Weak interoperability is probably the most notable  😔

jmacura:
Some more work needs to be done. Simple bug-fixes and enhancements forked into #2100

jmacura:
The test case provided by @raitisbe seems to work OK now. I have re-refactored some of the code back to its original form (`layerDimensionDefinitionChanges` expects `Layer` again instead of `HsLayerDescriptor`)
image

jmacura:
> 
> 
> Got some text overlaying the map when the sidebar is on the left side and moving mouse over the time slider
> image

I can't replicate that one...
image

PR #1998, opened 2021-07-14
Versions update, Layer manager extension

SLeitgeb:


package.json
```diff
@@ -100,6 +100,7 @@
     "karma-jasmine-html-reporter": "^1.7.0",
     "prettier": "^2.3.2",
     "ng-packagr": "^11.2.4",
+    "prettier": "^2.3.1",
```
raitisbe, line 103: Prettier seems to be listed twice

projects/hslayers-material/package.json
```diff
@@ -3,18 +3,18 @@
...
+    "@angular/cdk": "^11.2.13",
+    "@angular/common": "^11.2.13",
+    "@angular/core": "^11.2.13",
+    "@angular/forms": "^11.2.13",
+    "@angular/localize": "^11.2.13",
+    "@angular/material": "^11.2.13",
+    "ol": "^6.5.0",
     "ol-popup": "^4.0.0",
-    "hslayers-ng": "^3.0.0"
+    "hslayers-ng": "^4.0.1"
```
raitisbe, line 15: Can target 5 now

raitisbe (changes requested):

SLeitgeb:
Added an option to specify layout widths with percentage values in d05b2ecbd5861b96145bc789dff7edb6c4bc7d77

raitisbe (approved):

raitisbe:
Thank you. Just a quick thing before merge, could you rename the commits and squash some of them if possible to follow https://www.conventionalcommits.org/en/v1.0.0/ standard?

SLeitgeb:
This should be much better hopefully! Thanks for the link, I was wondering where those neat commit messages came from :)

raitisbe:
> This should be much better hopefully! Thanks for the link, I was wondering where those neat commit messages came from :)

You can also create the commit with `npm run commit` using a wizard ;)

PR #1999, opened 2021-07-14
Hide sidebar for multiple HsConfig settings

DailisLangovskis:
Or componentsEnabled.guiOverlay is set to false
Or componentsEnabled.sidebar is set to false


FilipLeitner (approved):
What about base map gallery? I suppose it should also be hidden on componentsEnabled.guiOverlay = false, right? I wasnt exactly sure so havent added it yet. Otherwise works good

PR #2018, opened 2021-07-26
fix: Read SLD style file for shp

### LINKED ISSUE #2015: Login button for Add data from file
When adding Shapefile and the user is not logged in to the catalogue, the "Add" button should not be enabled, but probably replaced by "Login to catalogue" button.

FilipLeitner:
+ GUI improvements for adddata shp

fixes #2015


fzadrazil (changes requested):
I would restrict the SLD input to the *.sld file extension.

FilipLeitner:
> I would restrict the SLD input to the *.sld file extension.

Sure, thats good point. Added change

fzadrazil (approved):

PR #2022, opened 2021-07-26
refactor: Improve WMS layers table UI

### LINKED ISSUE #2008: Layer titles unreadable in the add-data layer list
..due to long names and un-scrollable content

image

FilipLeitner:
fixes #2008

### DIFF
--- a/projects/hslayers/src/components/add-data/url/wms/add-data-url-wms.service.ts
+++ b/projects/hslayers/src/components/add-data/url/wms/add-data-url-wms.service.ts
@@ -179,6 +179,13 @@ export class HsAddDataUrlWmsService {
 
       this.data.services = this.filterCapabilitiesLayers(caps.Capability.Layer);
 
+      //Make sure every service has a title to be displayed in table
+      for (const service of this.data.services) {
+        if (service.Title.length == 0) {
+          service.Title = service.Name;
+        }
+      }
+
       //TODO: shalln't we move this logic after the layer is added to map?
       if (layerToSelect) {
         const serviceLayer = this.data.services.filter(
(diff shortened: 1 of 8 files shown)


jmacura:
Nice improvement, as a simple fix it's great.
image

I suggest a bit more complex improvement: additionally to these changes, add (click) action to each line, which would expand the text in full length (all cells, single line). What do you think?

raitisbe:
The name should be somewhere in case the titles for 2 layers are the same, but names differ. Name as hover tooltip maybe?

FilipLeitner:
> The name should be somewhere in case the titles for 2 layers are the same, but names differ. Name as hover tooltip maybe?

Hmm I mean now 

> The name should be somewhere in case the titles for 2 layers are the same, but names differ. Name as hover tooltip maybe?

Added name to the tooltip. It feels a bit tricky but works for me on Chrome, Firefox and also Edge so I hope itll be alright.

raitisbe (approved):

projects/hslayers/src/components/add-data/url/wms/add-data-url-wms.service.ts
```diff
@@ -100,6 +100,18 @@ export class HsAddDataUrlWmsService {
     this.getDimensionValues = hsDimensionService.getDimensionValues;
   }
 
+  expandTableRow(e) {
+    const action = e.target.className.includes('hs-wms-expandedRow')
```
raitisbe, line 104: Was there no angularish way to do this?

raitisbe (approved):

projects/hslayers/src/components/add-data/url/wms/add-data-url-wms.service.ts
FilipLeitner: What exactly?

raitisbe (approved):

PR #2028, opened 2021-07-28
Stick ol to version 6.5 for a while

jmacura:
This is temporarily preserving from upgrade to @openlayers >= 6.6, as it is creating a throng of errors during build.


raitisbe (approved):
Agree

PR #2051, opened 2021-08-04
Issue 2045/upload to existing

### LINKED ISSUE #2045: Add data (upload file) panel shoud support uploading features to an existing layer

FilipLeitner:
closes #2045

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "@ng-bootstrap/ng-bootstrap": "^9.1.3",
     "@ngx-translate/core": "^13.0.0",
     "@ngx-translate/http-loader": "^6.0.0",
-    "@tmcw/togeojson": "^4.4.1",
     "@types/cesium": "^1.67.14",
     "bootstrap": "^4.6.0",
     "cesium": "^1.84.0",
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "@ng-bootstrap/ng-bootstrap": "^9.1.3",
     "@ngx-translate/core": "^13.0.0",
     "@ngx-translate/http-loader": "^6.0.0",
+    "@tmcw/togeojson": "^4.4.1",
     "@types/cesium": "^1.67.14",
     "bootstrap": "^4.6.0",
     "cesium": "^1.84.0",
--- a/projects/hslayers/src/components/add-data/vector/add-data-vector.component.ts
+++ b/projects/hslayers/src/components/add-data/vector/add-data-vector.component.ts
@@ -114,7 +114,7 @@ export class HsAddDataVectorComponent implements OnInit {
    */
   async add() {
     this.uploadType == 'new'
-      ? this.addNewLayer()
+      ? await this.addNewLayer()
       : await this.updateExistingLayer();
 
     this.hsLayoutService.setMainPanel('layermanager');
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -42,6 +42,7 @@
     "@ng-bootstrap/ng-bootstrap": "^9.1.3",
     "@ngx-translate/core": "^13.0.0",
     "@ngx-translate/http-loader": "^6.0.0",
+    "@tmcw/togeojson": "^4.4.1",
```
raitisbe, line 45: Couldn't we use the already exisiting OL parsers for gpx and kml?

FilipLeitner: Well, ye it has got to be possible, sure. Thanks. I'll rewrite it .. got this part from Dailis as he had that implemented for whiteboard

projects/hslayers/src/components/add-data/vector/add-data-vector.service.ts
```diff
@@ -276,16 +257,44 @@ export class HsAddDataVectorService {
       console.log('Uploaded file is not supported!');
     }
   }
+
+  async readUploadedFile(file: any): Promise<any> {
+    let uploadedData: any = {};
+    const fileType = this.tryGuessTypeFromNameOrUrl(file.name.toLowerCase());
+    switch (fileType) {
+      case 'kml':
+        uploadedData = await this.createVectorObjectFromXml(file, fileType);
```
DailisLangovskis, line 266: @FilipLeitner The only problem with this is that You cannot get features array from xml, as they are included inside the base64 url, with the styling.

DailisLangovskis (commented):
Otherwise everything works great, as far as I could test it.

projects/hslayers/src/components/add-data/vector/add-data-vector.service.ts
FilipLeitner: Ye feature count doesnt work with kml but it could be read with kml parser. Should probably add it to not keep it missleading. But this approach is used only when adding KML 'as a whole' . Otherwise, when adding kml features into existing layer 'convertUploadedData' data function is used. Should work.

DailisLangovskis: Yes it works. Kinda doesn't feel right parsing data just for the feature count to pop up, but it must be done.

FilipLeitner: Ye thats why I havent done it right away. I could maybe just change description for KML to not include the number. What do you think?

DailisLangovskis: Agree on this, but lets also hear what others think of this idea.

projects/hslayers/src/components/add-data/vector/add-data-vector.component.ts
```diff
@@ -82,10 +98,41 @@ export class HsAddDataVectorComponent {
...
+
+  async updateExistingLayer(): Promise<void> {
+    let features = this.features.length > 0 ? this.features : [];
+    if (this.dataType != 'geojson') {
+      const kml = await this.hsAddDataVectorService.convertUploadedData(
+        this.vectorFileInput.nativeElement.files[0]
+      );
+      features = kml.features; //proper typing will get rid of this
+    }
+    this.sourceLayer.getSource().addFeatures(features);
```
raitisbe, line 132: Could you check if this is clustered layer and add to getSource().getSource() in that case?

projects/hslayers/src/components/add-data/vector/add-data-vector.service.ts
raitisbe: I think its ok to parse before adding the layer. This way we might catch parsing errors before some broken layer gets added to map.

raitisbe (changes requested):

raitisbe:
Thank you. I tested on whiteboard and it works great!

raitisbe (approved):

PR #2056, opened 2021-08-05
feat: Support downloading of queried feature as GeoJSON

raitisbe:


FilipLeitner:
Would be nice to be able to download multiple selected features as one file... Might add save button next to a heading or something simillar.

raitisbe:
> Would be nice to be able to download multiple selected features as one file... Might add save button next to a heading or something simillar.

Right, I didnt even realise we can select multiple features :)

raitisbe:
Creating a followup issue https://github.com/hslayers/hslayers-ng/issues/2062

PR #2069, opened 2021-08-08
fix: Properly select and add layer loaded through url param 'wms_layers'

FilipLeitner:
Little change that fixes loading layers from url


raitisbe (approved):

raitisbe:
Thanks

Issue #2111, opened 2021-08-30
Decouple components


raitisbe:
# Feature Request

Hslayers components are not tree-shakable and all of them are included in every application which uses hslayers. For example trip-planner, sharing and location, add-data searching functionality might not always be needed, but still are included in all applications.  Add data is especially heavy because it imports large OL modules (KML support).

- We should reduce the cross importing of hslayers modules (services)
- HsLayoutComponent and its PanelSpace should be dynamically populated with components from container application, similarly as with dialog modals and toasts.
- Move nonessential module imports out of HsCoreModule into HsLayersModule which would contain everything
- Make importing of HslayersModule optional giving container application developers the  possibility to import HsLayers modules separately similarly how it is done with NgBootstrap - NgbModule vs NgbPaginationModule (https://ng-bootstrap.github.io/#/getting-started#imports)
- Just an afterthought about HsEventBus . Its actually good to use event bus for this reason, because it reduces the interdependency of HsLayersModules, even though the event Subjects are not grouped by function, but by form. If we want to have some decoupling between modules still, but don't want to use HsEventBus, then we can split the RxJs Subjects out of the particular modules/services into small files having just the subjects exported as constants. In that case we can import those and not import the Hs.. Angular module with quite many components and services for some of the panels.

jmacura:
https://angular.io/guide/lightweight-injection-tokens

raitisbe:
> https://angular.io/guide/architecture-services#providing-services
> https://angular.io/guide/lightweight-injection-tokens

Very interesting, even if we are not using @ContentChild that much. Wonder if those tokens would work for such dynamic components such as dialogs or panels (later)

jmacura:
> 
> 
> > https://angular.io/guide/architecture-services#providing-services
> > https://angular.io/guide/lightweight-injection-tokens
> 
> Very interesting, even if we are not using @ContentChild that much. Wonder if those tokens would work for dynamic components such as dialogs or panels (later)

Aye, it won't affect us that much now, as I can't recall we use @ContentChild anywhere.
Meanwhile I found this Medium article which seems to be much more relevant to us: https://tomastrajan.medium.com/the-best-way-to-architect-your-angular-libraries-87959301d3d3

PR #2131, opened 2021-09-06
Public API + refactoring

jmacura:
------------

## Description
Multitude of wide refactoring changes related (more or less) to #2111 

#2111 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [x] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

## Checklist


jmacura:
All our services are provided in 'root' by default, i.e. they are not coupled with any particular module.
Only exceptions are:
* EndpointsWithDatasourcesPipe which is a Pipe also used as a service inside another service (sounds horrible but works)
* HsConfig
* HsLayerUtilsService
* HsUtilsService
the latter three can be probably also provided globally, but for now I have kept the status quo.

jmacura:
On the exports. The primary public-api was split into separate public-api files in each sub-module for easier maintanance. Now everything is export for public. All
* components
* modules
* services
* pipes
* interfaces
* plain classes and
* types.

Only spec-files and mock-files are not exported + `backwards-compatibility.ts` file from styler.

raitisbe:
Can we merge this as soon as possible? it seems to be quite safe from the first glance at the changes and also tests are completing.

jmacura:
Well, yes, I would just like to hear your opinion about the public exports. Whether it is OK to expose 100% of the code to the consumers, or whether we want to keep some services/pipes/types/anything for internal use only. @raitisbe @fzadrazil perhaps @FilipLeitner @DailisLangovskis but can be also important for @SLeitgeb @ChcJohnie ..?

Beside that I am fine with merging, we can do more refinement later on.

DailisLangovskis:
> Well, yes, I would just like to hear your opinion about the public exports. Whether it is OK to expose 100% of the code to the consumers, or whether we want to keep some services/pipes/types/anything for internal use only. @raitisbe @fzadrazil perhaps @FilipLeitner @DailisLangovskis but can be also important for @SLeitgeb @ChcJohnie ..?
> 
> Beside that I am fine with merging, we can do more refinement later on.

I have no problem with giving 100% to public! :)

raitisbe:
> Well, yes, I would just like to hear your opinion about the public exports. Whether it is OK to expose 100% of the code to the consumers, or whether we want to keep some services/pipes/types/anything for internal use only. @raitisbe @fzadrazil perhaps @FilipLeitner @DailisLangovskis but can be also important for @SLeitgeb @ChcJohnie ..?
> 
> Beside that I am fine with merging, we can do more refinement later on.

Theres nothing secret. The only rationalle was to not overload the mental capabilities of consumers of the API :) Also we would need to document everything we expose at some point, but yes it's safer to expose everything not to forget something.

raitisbe (approved):

jmacura:
Great, I also support the publicity :-) Yet it's worth the note that it has some side-effect for us: When everything is part of a public API, then every renamed/deleted/moved function, method or even class property variable shall be considered a breaking change 😵

raitisbe:
> Great, I also support the publicity :-) Yet it's worth the note that it has some side-effect for us: When everything is part of a public API, then every renamed/deleted/moved function, method or even class property variable shall be considered a breaking change

Can't we reexport the lower level public_apis into root 'hslayers-ng' public api thus making it nonbreaking? Or does it somehow interfere with the treeshaking? I also noted that inside our modules when using import {HsPanelComponent} from '../layout/public-api' it created a circular reference, but a more speciffic `import {HsPanelComponent} from '../layout/panels/panel-component.interface'`; was working fine. Must be carefull with the automatically generated imports.

jmacura:
> 
> 
> > Great, I also support the publicity :-) Yet it's worth the note that it has some side-effect for us: When everything is part of a public API, then every renamed/deleted/moved function, method or even class property variable shall be considered a breaking change
> 
> Can't we reexport the lower level public_apis into root 'hslayers-ng' public api thus making it nonbreaking?

This is not a problem at all. For a consumer, the internal structure of our repo is obscured. As a consumer, you just `import {HsPanelComponent} from 'hslayers-ng'` no matter in which sub-directory it resides. The problem is, when you make an app where you have code like
```
import {HsPanelContainerService} from 'hslayers-ng'
...
constructor(
    public hsPanelContainerService: HsPanelContainerService,
  )
...
this.hsPanelContainerService.destroy(somePanel);
```
and then we suddenly decide that `demolish()` could be a better naming then `destroy()`, so we rename, and the app is broken. The same happens when we move the `destroy()` method into some other service or just rename the service. Dumb example perhaps but I guess you get the point 😉

Issue #2150, opened 2021-09-10
Legend panel does not reflect all symbolizer changes in styles


jmacura:
# Bug

**Describe the bug**
If fill outline is set in Fill symbolizer, it overwrites the line colour of Line symbolizer in Legend panel. Legend also displays no symbols for markers (Mark symbolizers).

**To Reproduce**
1. Set custom styles for some layer: change colour and width of lines and/or polygons.; set some marker
2. Go to Legend panel
4. See default styles

**Expected behavior**
As in the following screenshot: Line symbolizer has purple colour set, Fill symbolizer has green outline set -> line shall appear purple in legend, polygon shall appear with green outline. Marker symbols shall be displayed the same way as Icon symbols are.

**Screenshots**
![image](https://user-images.githubusercontent.com/5598693/132927171-3fe20127-90f3-4e13-9219-2f448b601f2c.png)
![image](https://user-images.githubusercontent.com/5598693/132927067-e9d4e2f8-83b2-4b1f-8828-5cde7eaf681c.png)

**Additional context**
Relates to #2148

fzadrazil:
Markers are not reflected in legend.
![image](https://user-images.githubusercontent.com/48727502/134700427-3111e5ef-7a3c-4966-b8fc-4743ee2a1d79.png)

raitisbe:
I think we need to move it it milestone 7.0 , since fix for this involves new peer dependencies meaning breaking change.

Issue #2148, opened 2021-09-10
Fill symbolizer affects line symbolizer (and vice versa)


jmacura:
# Bug

**Describe the bug**
When a Line symbolizer is set, it also affects the symbology of polygon boundaries. Vice versa, when a Fill symbolizer is set with some outline colour, it affects the symbology of lines.

**To Reproduce**
1. Open some layer with polygons and lines or draw some.
2. Go to symbolize layer
3. Create Line symbolizer
4. Set some colour and width

**Expected behavior**
Line symbolizer shall be only applied to lines. Fill symbolizer shall be only applied to polygons.

**Screenshots**
![image](https://user-images.githubusercontent.com/5598693/132926342-321d573d-3edb-4599-a527-ac9472ce2a02.png)
![image](https://user-images.githubusercontent.com/5598693/132926360-e8b6612b-b8fc-4a96-a96a-81c06b0fa969.png)
![image](https://user-images.githubusercontent.com/5598693/132926369-0a53a6bd-ba19-4cb4-a6fd-0ef84d768c5d.png)

**Additional context**
Or perhaps not... Could this be a desired behaviour?

raitisbe:
Yeah, I dont think we can classify it as a bug. User must know what he's doing and probably the fix would need to be made in the geostyler anyway. Will leave it open for a bit.

fzadrazil:
What is misleading here is the fact that there is an outline settings in the Fill symbolizer and line settings in Line symbolizer which does the same, but are not synchronized. So the user would suppose Outline only affects fill symbols and Line symbolizer only polylines. Eg. when you change the outline color, it won't change the value in Line symbolizer, but the lines get changed and vice versa.

raitisbe:
Yes, but looking from the other side, there would not be a possibility to draw these nice contours around the line in another color, if we synchronized fill symbolizers outline color to line symbolizers color.

jmacura:
Just leaving a note that this will be confusing for any end-user and I still consider it a UX-bug.
![obrazek](https://user-images.githubusercontent.com/5598693/156794117-3067bef7-db02-4fe9-88db-cbb090aa95e2.png)

PR #2185, opened 2021-09-20
Issue 2183

### LINKED ISSUE #2183: Enable reordering symbolizers
# Feature Request

Order of symbolizers affects the resulting symbol. Eg. outline color of fill symbolizer can completely cover polyline if line symbolizer is defined after the fill.

Enable to change order of symbolizers, ideally with drag&drop. It would be also nice to be able to collapse/expand each symbolizer panel.

DailisLangovskis:
## Description

Introduces drag and drop feature for styler rules and symbolizers

closes #2183 #2150 
refs #2148 

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist


DailisLangovskis:
@FilipLeitner I would appreciate if You could add the missing translations as well as, if it is possible to fix the symbolizer type dropdown, that is getting opened behind other DOM elements, couldn't get it right.

projects/hslayers/src/components/styles/styles.module.ts
```diff
@@ -1,10 +1,13 @@
+import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
 import {CUSTOM_ELEMENTS_SCHEMA, NgModule} from '@angular/core';
 import {CommonModule} from '@angular/common';
+import {DragDropModule} from '@angular/cdk/drag-drop';
 import {FormsModule} from '@angular/forms';
-import {TranslateModule} from '@ngx-translate/core';
+import {MatExpansionModule} from '@angular/material/expansion';
```
raitisbe, line 6: I dont like to mix material design into our bootstrap version of hslayers. Becomes too much of a Frankenstein.

DailisLangovskis: I understand, but couldn't get drag and drop to work on ngb

DailisLangovskis: I will try to make it work with ngb then.

DailisLangovskis: How ever we could consider to move to angular-material at some point and remove ngb dependencies. (Just a thought)

raitisbe: I think the correct way would be to have a UI kit agnostic hslayers core and then have 2 more extensions/libraries - one for NgBootstrap one for Material

FilipLeitner:
> @FilipLeitner I would appreciate if You could add the missing translations as well as check the css and give me a notice on some errors, feel free to give me some suggestions. Also when adding the styler.component.scss file as import to global app.scss, it causes some build error, coming from overwrite functions.

Sorry I must have missed the @ mention email notification among all those dependency upgrade ones so I didnt even know you that u asked for anything. Will do ! :)

raitisbe:
The correct conventiopnal commits prefix is `feat` instead of `feature` ;)

FilipLeitner:
btw :D I spent some quality time trying to get that styling to work some other way  but because of the nested cdk dnd it might as well stay as it is.
- compilation error is bacuase of unexpected value being passed to to-hsl function (rgba definition not strarting with 0) . Could be solved by coverting rgba to hsla. If thats done and styler.component.scss declarations are removed from styleUrls we need to style dnd preview ourselves. Its a bit tricky as it gets inserted inside body and styles not getting passed to children + more.

...so long story short Id leave it as it is (Y) . ill try to provide the  translations at least :P

FilipLeitner:
Wonder where could the 'Reordering' toggle button go. Its not really convenient up there in my opinion. Had to go through code to figure it out :D

raitisbe:
I think we don't need the button at all and reordering can be toggled on by default.

DailisLangovskis:
> The correct conventional commits prefix is `feat` instead of `feature` ;)

Sorry, my bad. Will reword it

raitisbe (approved):

FilipLeitner (approved):

jmacura:
The drag'n'drop feature is really cool, but a bit eager. It starts dragging when I click&move the mouse button _anywhere_ in the styler. It means, when I want to play with the range buttons like Opacity or Width, it starts dragging instead...
image

DailisLangovskis:
> The drag'n'drop feature is really cool, but a bit eager. It starts dragging when I click&move the mouse button _anywhere_ in the styler. It means, when I want to play with the range buttons like Opacity or Width, it starts dragging instead...
> image

Well that is why I had this toogle for dragging, to avoid it, that I removed on the request. Must admin, didn't notice it when using sliders.

jmacura:
This PR does not fully resolve #2150 . Markers are not reflected in legend.
image

jmacura:
> 
> Well that is why I had this toogle for dragging, to avoid it, that I removed on the request. Must admin, didn't notice it when using sliders.

Is it not possible to make the block draggable only by its header?
This I mean: 
image

DailisLangovskis:
> > Well that is why I had this toogle for dragging, to avoid it, that I removed on the request. Must admin, didn't notice it when using sliders.
> 
> Is it not possible to make the block draggable only by its header?
> This I mean:
> image

Well the thing is that tha cdkDrag item is a component, not a one div or smth. Had to create a component list item, because there are no uniq identifiers inside styleOnject rules or symbolizers I could relay on the check which rule was opened and whick symbolizer.

DailisLangovskis:
> > > Well that is why I had this toogle for dragging, to avoid it, that I removed on the request. Must admin, didn't notice it when using sliders.
> > 
> > 
> > Is it not possible to make the block draggable only by its header?
> > This I mean:
> > image
> 
> Well the thing is that tha cdkDrag item is a component, not a one div or smth. Had to create a component list item, because there are no uniq identifiers inside styleOnject rules or symbolizers I could relay on the check which rule was opened and whick symbolizer.

In this case the answer would be no.

raitisbe:
> > > > Well that is why I had this toogle for dragging, to avoid it, that I removed on the request. Must admin, didn't notice it when using sliders.
> > > 
> > > 
> > > Is it not possible to make the block draggable only by its header?
> > > This I mean:
> > > image
> > 
> > 
> > Well the thing is that the cdkDrag item is a component, not a one div or smth. Had to create a component list item, because there are no uniq identifiers inside styleObject rules or symbolizers I could relay on to check,which rule was opened and which symbolizer.
> 
> In this case the answer would be no.

Could search a bit about event bubbling and clickthrough, maybe something comes up. It is possible to handle mouse events only in nested components, but not on the parent.

raitisbe:
Also: https://stackoverflow.com/questions/57710285/set-a-part-of-a-draggable-div-as-undraggable

DailisLangovskis:
> This PR does not fully resolve #2150 .
> 
> Markers are not reflected in legend.
> image
> 
> Neither the Fill-border vs. Line clash is addressed.
> image

The clash gets resolved by draging symbolizer order, the only way to do it.

DailisLangovskis:
> This PR does not fully resolve #2150 .
> 
> Markers are not reflected in legend.
> image

This must be addressed in a seperate PR, because it is not an easy small fix to implement!

DailisLangovskis:
> The drag'n'drop feature is really cool, but a bit eager. It starts dragging when I click&move the mouse button _anywhere_ in the styler. It means, when I want to play with the range buttons like Opacity or Width, it starts dragging instead...
> image

Resolved!

fzadrazil:
Reordering of symbolizers works great now, thank you. Just one more minor request. When I add a new symbolizer, the dropdown list with symbolizer types stays open now. It should close automatically. Testing this in the hslayers/test-app project.

fzadrazil (approved):

PR #2315, opened 2021-10-21
Dailis fixes for extent feature objects and external layer and composition objects

DailisLangovskis:
## Description

This PR fixes unwanted self looping objects. Extent features, that includes feature, that includes the same feature over and over and also the same for external layers or compositions, including endpoint that includes the same layers or compositions again. Solution is to use id and reference it where needed.

Could use deeper testing, because I am not 100% sure I have tested all possible scenarios.

None

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/common/feature-extensions.ts
+++ b/projects/hslayers/src/common/feature-extensions.ts
@@ -50,14 +50,6 @@ export function getFeatureName(feature: Feature<Geometry>): string {
   return feature.get(NAME);
 }
 
-export function setRecord(feature: Feature<Geometry>, record: any): void {
-  feature.set(RECORD, record);
-}
-
-export function getRecord(feature: Feature<Geometry>): any {
-  return feature.get(RECORD);
-}
-
 export function setFeatures(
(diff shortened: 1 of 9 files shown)


projects/hslayers/src/common/endpoints/endpoint.interface.ts
```diff
@@ -15,6 +15,7 @@ export interface HsEndpoint {
   type: string;
   title: string;
   url: string;
+  id: string;
```
raitisbe, line 18: Not sure I like this mandatory attribute.

DailisLangovskis: I just thought using Id would be a good way of referencing the endpoint, rather then adding it to each layer over and over.

raitisbe:
From memory point of view object reference is cheaper then storing string
variable

PR #2402, opened 2021-11-16
Geometric operations in Drawing

### LINKED ISSUE #2386: Geometric operations in Drawing
# Feature Request

Add more tools to drawing for polygons like

- split polygon by polyline
- union
- intersection
- difference
- (symetrical difference)?

FilipLeitner:
## Description

- Implement geometric operations(difference, union, intersect, 'cut by line') to drawing panel
- Some perf improvements

closes #2386 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist


FilipLeitner:
> @FilipLeitner Inside the edit panel, using split and having selected at least 2 feature, I am getting this error -> err
> 
> P.S Added lv translations

Thanks! For translation and also for a quick test. It still needs some fine tuning for sure.

jmacura:
After some testing of the "splitting" tool, I have reached the point where two polygons seemed to be split, but were clickable only as a whole and the editing panel was not showing anything.
image

There were several errors in the console.
image
etc.

FilipLeitner:
Thanks for the translations and testing @jmacura !  Hopefully ill be able to replicate and fix

FilipLeitner:
Added few new fixes  but I could not replicate the bug u found @jmacura .

jmacura:
> Added few new fixes but I could not replicate the bug u found @jmacura .

Try to split the polygon and union it again. I always end up with some unexpected results after several repetitions.
image

jmacura:
Ah, I think this has something to do with snapping. When I draw the split line with as snapped to one of the corners, it actually does not take the coordinates of the corner, it only visually "snaps" the line, but it is not really intersecting the polygon.

image
image

jmacura:
And is it possible to disable editing of other features when drawing the intersection features? It is annoying to move vertices of the underlaying polygons when you actually want to draw something over them.
image

FilipLeitner:
Trying to union features that have been split is not really possible right now because there is actually a gap between the split parts of the original geometry
image
even after using 0.001 meterr as a line buffer distance (the edege of the polygon seems ti be visually bugged becuase moving a single vertex smoothen  the line immediately. Also there are no vertexes for each of the 'bumps'
image

It would be possible to create multipolygon with the little gap but that complicates everything....other solution that came to my mind is to look for a way to buffer features that user want to union by the same distance we 'buffer' splitting line. Maybe even

You are right about the snap intearction.  Ill try to turn it off same with the editing

raitisbe:
> Trying to union features that have been split is not really possible right now because there is actually a gap between the split So if I understand correctly the split polygons are not topologically correct i.e. don't have a common point?

FilipLeitner:
> > Trying to union features that have been split is not really possible right now because there is actually a gap between the split So if I understand correctly the split polygons are not topologically correct i.e. don't have a common point?

Exactly

raitisbe (approved):

Issue #2420, opened 2021-12-03
Load QML symbology for WFS in compositions


fzadrazil:
# Feature Request

**Is your feature request related to a problem? Please describe.**
QML symbology is loaded correctly for single WFS layers, but not for WFS included in map compositions.

**Describe the solution you'd like**
Implement loading of QML symbology for layers contained in compositions as well as for single WFS layers.

PR #2422, opened 2021-12-03
fix: Parse style when loading WFS from compositions

raitisbe:
## Description

Parse style from composition when loading WFS layers

#2420 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/add-data/catalogue/layman/layman.service.ts
+++ b/projects/hslayers/src/components/add-data/catalogue/layman/layman.service.ts
@@ -280,17 +280,7 @@ export class HsLaymanBrowserService {
     const lyr = await this.fillLayerMetadata(ds, layer);
     let style: string = undefined;
     if (lyr.style?.url) {
-      try {
-        style = await this.http
-          .get(lyr.style?.url, {
-            headers: new HttpHeaders().set('Content-Type', 'text'),
-            responseType: 'text',
-            withCredentials: true,
-          })
-          .toPromise();
-      } catch (ex) {
-        console.error(ex);
-      }
+      style = await this.getStyleFromUrl(lyr.style?.url);
     }
     if (lyr.style.type == 'sld') {
       if (!style?.includes('StyledLayerDescriptor')) {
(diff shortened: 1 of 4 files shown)


raitisbe:
This should in theory solve it, I was just not successfull in finding any composition with a QML style in the composition definition to test on.

fzadrazil:
You can try this composition, it still fails to load the style for its layers https://hub.lesprojekt.cz/rest/workspaces/fzadrazil/maps/demo2/file

raitisbe:
yes, because `style` attribute is not set anywhere in the composition json->layers items. And https://hub.lesprojekt.cz/geoserver/fzadrazil/wfs points to wfs service which is XML.

It's different when we are adding layers from layman directly, since there we know the style url.

raitisbe:
@fzadrazil Could it be that these compositions containig wfs were created in qgis? Because we are setting style attribute on layer in composition json when saving.

fzadrazil:
> @fzadrazil We need to set the style attribute for layer in composition in qgis. Because in HSL we are setting style attribute on layer in composition json when saving.

Yes, I am just thinking where to put the style url in the composition schema. There is a params.styles property, but that is, I think, meant only for WMS style name. And property "style" is for explicit definition of style, not link to sld/qml. We could propose something in the context of this [Map Composition PR](https://github.com/hslayers/map-compositions/pull/1), eg. change the data type of "style" property so it is not an object, but could be also a string (url)?

```
"style": {
  "$id": "#/properties/layers/items/properties/style",
  "type": ["object", "string"],
  "title": "Style definition of vector layer features or url to style file (QML, SLD)",
  ...
}
```

raitisbe:
> 

Yes, definitely. It thought it already is a string, since I'm storing SLD directly there. Object type was meant for the old way of storing styles pre-SLD era.
I can also check if the content begins with `http: ` and in that case do another request to fetch the style from server.

raitisbe:
@fzadrazil Yup, I just checked it already is string in laymans version of the schema at https://github.com/LayerManager/layman/blob/master/src/layman/map/schema.draft-07.json
image
Seems, that I have forgotten to update it in our wiki and components/compositions/schema.json
Too many places..

And yes, we can't use WMS params STYLE.

fzadrazil (commented):
I have requested the change in QGIS plugin/Layman. Once it is implemented, I will test again in hslayers.

raitisbe:
It should load the style contents from a file defined in style: attribute of layer in composition.

jmacura:
FTR: This feature is unit-tested but not live-tested in any real-world scenario. The PR was merged nevertheless in an attempt to be "future-proof". Once the feature is implemented in the QGIS Layman plugin, the functionality shall be retested.

PR #2458, opened 2021-12-15
Map swipe (Issue 2421)

### LINKED ISSUE #2421: Map swipe control panel
# Feature Request

Create new tool which will introduce a map swipe control so the user is able to compare two map contents while the view is synchronized. Eg. https://viglino.github.io/ol-ext/examples/control/map.control.swipe.html
https://github.com/Viglino/ol-ext/blob/master/src/control/Swipe.js
There can be several options how to define the map content of both the windows. That is for a discussion:

- each window displays just one layer, there are two dropdowns with the list of available layers
- left window shows the default map of the application / right window shows one selected layer (dropdown list in the panel)
- content of both windows is defined by checkbox lists of available layers

I guess the selected solution depends on what the swipe control component is capable of.

**Additional context**
One existing application for inspiration https://app.iprpraha.cz/apl/app/prahavcera/
But we would like to have more general tool for HSLayers, not just one specific application.

DailisLangovskis:
## Description

This PR introduces a new feature map swipe control and panel.

closes #2421 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist


DailisLangovskis:
@FilipLeitner Hey, If You have any idea, how the css rule hs-map-swipe-list min-height property could be set dynamically, based on the table body offsetHeight, that would be great. I tried using ViewChild and afterInitView add the property, but that cause the angular assertion error, value has changed after being checked, so I gave up. Maybe You have some tricks up in Your sleave.

DailisLangovskis:
@jmacura Could Your grammar skills regarding translations :P

projects/hslayers/src/common/layer-shifting/layer-shifting.service.ts
```diff
@@ -1,123 +1,78 @@
...
+import {HsEventBusService} from '../../components/core/event-bus.service';
+import {HsLayerManagerService} from '../../components/layermanager/layermanager.service';
+import {HsMapService} from '../../components/map/map.service';
+import {HsUtilsService} from '../../components/utils/utils.service';
+import {getBase} from '../layer-extensions';
+
+export class LayerListItem {
   title: string;
-  layer: Layer<Source>;
+  layer: any;
```
raitisbe, line 14: You can specify another type after  `|` instead of using any

DailisLangovskis: Couldn't use layer source because of the on listeners inside the swipecontrol class
.

raitisbe: What about using `any` there?

DailisLangovskis: @raitisbe Added change for the type to be Layer<Source>, but made layer to be (as any) inside the swipeControl class.

raitisbe:
Would be good to be able to use layer swiping also without extra panel - just through importing the swipe service.
Could also have a property in layer definition - `swipeSide: 1 | 2` , which would then distribute the layers to the necessary collections.

DailisLangovskis:
> Would be good to be able to use layer swiping also without extra panel - just through injecting the swipe service and calling some init method there. In simple app it would need be injected and executed by default. Could also have a property in layer definition - `swipeSide: 1 | 2` , which would then distribute the layers to the necessary collections.

Okey, will implement this.

jmacura:
> @jmacura Could Your grammar skills regarding translations :P

Let's see what I can do 😏

jmacura:
The `initialSwipeRight` config option is not reflected in the map-swipe panel.
Excerpt from the test-app:
```
initialSwipeRight: [polygons],
mapSwipeOrientation: 'vertical',
```
On load it's correct in the map
image

but not in the panel
image

btw, can we do something with the overflown text, @FilipLeitner ? 🥺

jmacura:
P.S. And the behaviour is tricky. When I do this, to "fix" the layer assignment...
image

it does this in the map...
image

DailisLangovskis:
> P.S. And the behaviour is tricky. When I do this, to "fix" the layer assignment... image
> 
> it does this in the map... image

There are two identical layers @jmacura

fzadrazil (changes requested):
It seems there is no way how to deactivate the map swiper now? Would be probably good to have the chance ;-)

DailisLangovskis:
> It seems there is no way how to deactivate the map swiper now? Would be probably good to have the chance ;-)

Do You mean like a button?

fzadrazil:
> > It seems there is no way how to deactivate the map swiper now? Would be probably good to have the chance ;-)
> 
> Do You mean like a button?

Well, the swipe feature is activated by opening the panel now, but no way how to turn it off. Toggle button is probably a good solution to activate/deactivate the swipe.

DailisLangovskis:
> > > It seems there is no way how to deactivate the map swiper now? Would be probably good to have the chance ;-)
> > 
> > 
> > Do You mean like a button?
> 
> Well, the swipe feature is activated by opening the panel now, but no way how to turn it off. Toggle button is probably a good solution to activate/deactivate the swipe.

Just have to figure out how to do it :D

raitisbe:
> 

call disableEvents function for all layers?

DailisLangovskis:
I tried to refactor the code, to use only one layers array, with property for wrapped layer swipeRight, but in the end I ended up with more lines of code and not working swipe control :D

jmacura:
The last changes make it awesome, but sometimes, when I am playing a really irresponsible user, I end up with a bloody broken map screen, like
image
or
image
or after some fiddling
image

There are no errors in the console. It's created by some combination of enabling the map-swipe and changing its direction. For some reason, only one half of the map is considered as swipe-able and gets split, which creates this fractal-like views...

jmacura:
The simplest "fix" might be just to disable the "horizontal swipe" button when the tool is disabled. Then no such obscurities could happen.

fzadrazil (approved):

fzadrazil:
Thank you Dailis for the latest updates. It now works nicely even in simple app and with the sidebar invisible. And the buttons for turning on/off look good like that.

jmacura (approved):
I very appreciate the numerous unit-tests :+1:
It looks and works perfectly.

raitisbe (approved):

PR #2479, opened 2021-12-22
Ability to change layer visibility in map-swipe (Issue-2478)

### LINKED ISSUE #2478: Allow to turn layers on/off from map swipe panel
# Feature Request

It might be too much to ask :-) But as I work with the swiper, it would be really great to have a chance to turn on and off the layers directly from the map swipe panel. Sth like that...
image

Do you think it is possible? If you have more layers in the map and need to switch between, it would be really handy not have to always switch to layer manager and back to map swipe.

**Alternative solution**
Another solution would be to display only currently visible layers in the map swipe list (there is no sense to display invisible layers there). The user would see then what layers are available for the swiper, but would still have to switch to layer manager to turn layers on and off.

DailisLangovskis:
## Description

This PR resolves issue 2478

closes #2478 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist


projects/hslayers/src/components/map-swipe/map-swipe.component.html
```diff
@@ -1,20 +1,30 @@
+<!-- Layer item template -->
+<ng-template #layerItem let-layer="layer">
+  <div class="hs-lm-item-title d-flex align-items-center" style="word-break:break-all; cursor:move">
+    <button type="button" class="btn btn-sm hs-lm-item-visibility m-0 p-0"
+      (click)="hsMapSwipeService.changeLayerVisibility(layer);$event.stopPropagation()"
+      [ngClass]="hsMapSwipeService.getLayerVisibility(layer) ? 'hs-checkmark' : 'hs-uncheckmark'"></button>
```
raitisbe, line 6: Not good to call getLayerVisibility function on each digest loop. Better to store visibility in some variable and update through observers.

fzadrazil:
It is perfect like this from the users perspective! Much more effective for work. So as soon as the remark from Raitis is solved, it is good to go from my side.

raitisbe (approved):

PR #2508, opened 2022-01-05
Ability to copy a layer (Issue 2506)

### LINKED ISSUE #2506: Ability to copy layer
# Feature Request

In regards to split map functionality would be good to be able to copy existing layer and then set different dimensions for that copy. To compare the same layer but in different time. 
Probably the best place for the copy button would be in layer editor before Delete.

DailisLangovskis:
## Description

Allows the user to create a copy of a layer from layer editor panel using a button and confirming the action.

closes #2506 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist


jmacura:
A nice enhancement to this would be: In the confirmation dialog, instead of simply asking for confirmation, ask for a new name for the copy. What do you think?

projects/hslayers/src/components/layermanager/layermanager.service.ts
```diff
@@ -1175,4 +1176,25 @@ export class HsLayerManagerService {
       window.URL.revokeObjectURL(url);
     }, 0);
   }
+
+  /*
+    Creats a copy of the currentLayer
+  */
+  copyLayer(newTitle: string): void {
+    const copiedLayer = new VectorLayer(
+      this.currentLayer.layer.getProperties()
```
raitisbe, line 1185: Will need to also clone the layers Source. Currently the cloned layer and original layer shares the same source object, which might create problems - changing dimension or other properties in one changes other.

raitisbe:
I think the Object.create() approach might create problems after all. I tried using that approach when cloning maps View and had unintended consequences - ol_uid was the same for the original and clone and the properties on clone such as center/zoom were still in sync with the original view when moving the map. I expect something similar might happen also with layer sources. Need to test more.

Bulletproof but a lengthy way would be to recreate the sources, copy getProperties() and serialize/deserialize features (maybe can also insert feature instances directly into the new source and they would be copied not referenced  - have to check) in case of vectorSource or copy the current layer params in case of WMS. `Object.create` sure looks tempting considering this.

DailisLangovskis:
> I think the Object.create() approach might create problems after all. I tried using that approach when cloning maps View and had unintended consequences - ol_uid was the same for the original and clone and the properties on clone such as center/zoom were still in sync with the original view when moving the map. I expect something similar might happen also with layer sources. Need to test more.
> 
> Bulletproof but a lengthy way would be to recreate the sources, copy getProperties() and serialize/deserialize features (maybe can also insert feature instances directly into the new source and they would be copied not referenced - have to check) in case of vectorSource or copy the current layer params in case of WMS. `Object.create` sure looks tempting considering this.

I will try to recreate the sources for each layer type

raitisbe:
Will need to get rid of owsFilling subject and rewrite to promises at some point

raitisbe (approved):

PR #2549, opened 2022-01-18
perf: Remove usage of panelSpaceWidth function

FilipLeitner:
## Description

Because panelSpaceWidth function is called on every digest cycle twice so I tought it might be quite beneficial for us to replace it with something more efficient which would be css. The problematic part is that the new approach must be able to handle dynamicaly created panel that are not part of the library which makes panel specific css classes unusable .
So I tought we might make a little trade off of absolute freedom of width declaration using for example 14 predefined 'size classes' generated like so:
```
    @for $i from 0 through 14{
      .hs-panelspace hs-panel-container .hs-panelWidth-#{400 + ($i * 25) } {
        width: 400 + ($i * 25)+px;
      }

      //Output:
      .hsl .hs-panelspace hs-panel-container .hs-panelWidth-400 {
          width: 400px;
        }
      .hsl .hs-panelspace hs-panel-container .hs-panelWidth-425 {
        width: 425px;
      }
      ....
```
These would be added to the panel element while being created -> based on `panelWidth` property of `data` argument in `hsPanelContainerService.create`  eg.:

- ` this.HsPanelContainerService.create({ component: AnyComponentClass });`

One problematic aspect I should point out is **how to find the right element** on which we add class onto. Basically looking for the element with `[hidden]=isVisible()`. However it doesnt necessarily need to be a first element (eg. query component) or div.
```
      componentRef.location.nativeElement.children[0].classList.add(
        `hs-panelWidth-${Math.round(panelItem.data.panelWidth / 25) * 25}`
      );
```
We cant use the host element of component because panel space would get the width from the most broad one.. 

So lets see maybe we can put it together somehow

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)
- [x] I am unsure (we'll look into it together)

- [x] Yes
- [ ] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/layout/layout.service.ts
+++ b/projects/hslayers/src/components/layout/layout.service.ts
@@ -85,7 +85,7 @@ export class HsLayoutService {
    */
   initializedOnce = false;
   /**
-   * Whether the app has been initialized already once. 
+   * Whether the app has been initialized already once.
    * Need this to not add panels wtice when NgRouter is used
    * @public
    * @default false
@@ -323,7 +323,7 @@ export class HsLayoutService {
       ows: 700,
       composition_browser: 550,
       addData: 700,
-      mapSwipe: 550
+      mapSwipe: 550,
     };
     const layoutWidth = this.layoutElement.clientWidth;
     Object.assign(panelWidths, this.HsConfig.panelWidths);
--- a/projects/hslayers/src/components/layout/panels/panel-container.component.ts
+++ b/projects/hslayers/src/components/layout/panels/panel-container.component.ts
@@ -14,6 +14,7 @@ import {HsPanelComponent} from './panel-component.interface';
 import {HsPanelContainerServiceInterface} from './panel-container.service.interface';
 import {HsPanelHostDirective} from './panel-host.directive';
 import {HsPanelItem} from './panel-item';
+import {HsConfig} from '../../../config.service';
 
 @Component({
   selector: 'hs-panel-container',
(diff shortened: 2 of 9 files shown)


DailisLangovskis:
@FilipLeitner Maybe the width panel param, could be optional and there could be a default value, if no width is provided?

FilipLeitner:
> @FilipLeitner Maybe the width panel param, could be optional and there could be a default value, if no width is provided?

`min-width: 400px` on panelspace class should acts as an fallback .
And about the width param, I was thinking about adding it as a standalone param but we would have two optionals then so I just added it inside existing data object.....but maybe with interface it would be also be more clear how to define the width.. Will change that

jmacura:
> So I tought we might make a little trade off of absolute freedom of width declaration using for example 14 predefined 'size classes'

A while ago (can be a year or more) we even discussed whether we should not remove this freedom for good and stick to uniform width for all the panels. Whereas every content that is too wide to be displayed in the panel shall be displayed in a dialog window. Just recalling what was said. I do not have strict opinion about it as of now.

raitisbe:
Freedom's good, don't we love freedom? :)

jmacura:
> Freedom's good, don't we love freedom? :)

We don't. That's why we use TypeScript and ESLint, to put an order in place ;-)

FilipLeitner:
After all mostly to prevent breaking changes some changes were made and panel width defintion will stay inside HsConfig (panelWidths)

raitisbe:
I have the problem that `componentRef.location.nativeElement.children` is empty, ebcause the nativeElement only contains one comment line `<!--container-->`. You can see it in hsl-7 branch of hslayers/applications AgroInfo.

image

FilipLeitner:
`<div *ngIf="isVisible()">` on that element. Dont remember exactly why I went with ngIf instead of just [hidden] but Id say it should be possible to use hidden. But ye if its empty then the class will not be added

raitisbe:
Lets have hs-panel-width- instead of hs-panelWidth- ?

raitisbe (approved):

PR #2575, opened 2022-01-21
feat: Add support for raster images with world file (Issue 356)

### LINKED ISSUE #356: Add support to load images with world file reference
Currently, local raster images can only be loaded by hard-typing their geographical extend. But that can be calculated from the associated world file (like .pgw, etc.) if provided.

DailisLangovskis:
## Description

Fixes issue #356.

Geotiff upload is now included under raster images

closes #356

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist


projects/hslayers/src/components/add-data/file/raster-images/raster-images.component.ts
```diff
@@ -25,7 +26,8 @@ export class HsFileGeotiffComponent
   }
   ngOnInit(): void {
     this.baseDataType = this.dataType;
-    this.acceptedFormats = '.tif, .tiff, .gtiff, .zip';
```
jmacura, line 28: `.gtiff` no longer accepted?

projects/hslayers/src/components/add-data/file/raster-images/raster-images.component.ts
```diff
@@ -8,13 +9,13 @@ import {HsLayoutService} from '../../../layout/layout.service';
 import {HsUploadedFiles} from '../../../../common/upload/upload.component';
 
 @Component({
-  selector: 'hs-file-geotiff',
-  templateUrl: 'geotiff.component.html',
+  selector: 'hs-raster-images',
```
jmacura, line 12: What about `hs-file-raster` or `hs-file-raster-image`?

projects/hslayers/src/components/add-data/file/raster-images/raster-images.component.ts
```diff
@@ -25,7 +26,8 @@ export class HsFileGeotiffComponent
   }
   ngOnInit(): void {
     this.baseDataType = this.dataType;
-    this.acceptedFormats = '.tif, .tiff, .gtiff, .zip';
+    this.acceptedFormats =
+      '.tif, .tiff, .tfw, .png, .pgw, .png.aux.xml, .jpeg, .jgw, .jpg.aux.xml, .jp2, .j2w, .zip';
```
jmacura, line 30: As per https://en.wikipedia.org/wiki/World_file#Filename_extension I would extend this to
`'.tif, .tifw, .tiff, .tiffw, .tfw, .png, .pngw, .pgw, .png.aux.xml, .jpg, .jpgw, .jpeg, .jpegw, .jgw, .jpg.aux.xml, .jp2, .jp2w, .j2w, .zip', .wld`

Also the common `.jpg` was missing.

jmacura (commented):
Nice improvement!
I have left some comments.
Translations are on the go 😉

projects/hslayers/src/components/add-data/file/file-type-values.ts
```diff
@@ -18,7 +18,7 @@ export const AddDataFileValues: Array<{id: AddDataFileType; text: string}> = [
     text: 'Shapefile',
   },
   {
-    id: 'geotiff',
-    text: 'GeoTIFF',
+    id: 'rasterImages',
+    text: 'Raster images',
```
jmacura, line 22: This shall be a i18n string as well

jmacura (changes requested):

projects/hslayers/src/components/add-data/file/raster-images/raster-images.component.ts
DailisLangovskis: @jmacura Well I checked layman expected types and gtiff is not included there, so I removed it

projects/hslayers/src/components/add-data/file/raster-images/raster-images.component.ts
DailisLangovskis: @jmacura Will include these changes, just wasn't sure if Layman supports them all, as they were wont mentioned in the [rest.md file](https://github.com/LayerManager/layman/blob/master/doc/rest.md#post-workspace-layers)

jmacura: @DailisLangovskis Oh, well, yes, then this shall be a feature request for the Layman team. @fzadrazil ?
Perhaps @jirik could give us some hint whether the list of extensions in their [REST documentation](https://github.com/LayerManager/layman/blob/master/doc/rest.md#post-workspace-layers) is exhaustive.

It would be of course unpleasant if we accepted something which would be refused on the Layman's side...

DailisLangovskis: .jpg was left out by accident

fzadrazil: > As per https://en.wikipedia.org/wiki/World_file#Filename_extension I would extend this to `'.tif, .tifw, .tiff, .tiffw, .tfw, .png, .pngw, .pgw, .png.aux.xml, .jpg, .jpgw, .jpeg, .jpegw, .jgw, .jpg.aux.xml, .jp2, .jp2w, .j2w, .zip', .wld`
> 
> Also the common `.jpg` was missing.

I think the list is accurate like this, but I honestly haven't tried all of them for the lack of sample data. I will confirm this with @jirik on Tuesday.

fzadrazil:
Are we able to detect the CRS of raster and prefill the textbox? When I try to upload raster in 3857, 4326 is entered by default and Layman fails to normalize it as a result. In case we are not able to do that, the CRS field should not be hidden in Advanced options and it would be better to leave it blank but mandatory.

projects/hslayers/src/components/add-data/file/raster-images/raster-images.component.ts
jirik: We tested only files mentioned in REST API documentation. Additional files .tifw, .tiffw, .pngw, .jpgw, .jp2w, .wld were not tested, but there is quite a chance Layman will accept it and gdal will detect georeferencing from them. I guess file .jpeg will not be recognized by Layman as main file currently.

DailisLangovskis: Updated the accesptedFormats predefined array

jmacura:
Some problems found during testing.

1) I have added two images with nearly the same extent and nearly the same name. But both is different. I can see them in the list of my layers in the Datasources panel. But only one can be added to the map. When I try to add the other, it looks like it was added, but it's still the first one.
image
image

jmacura:
2) There should be no option to add raster image as a WFS service.
image

jmacura:
3) When the SRS option is mandatory, it should not be hidden under "Advanced options". It is very unclear, why the "Add" button is disabled until the user opens the advanced options.
image

projects/hslayers/src/components/add-data/file/raster-images/raster-images.component.ts
jmacura: @DailisLangovskis @jirik @DailisLangovskis Tested now with `.pngw` and `.wld`. Both went fine 👍

DailisLangovskis:
> 3. When the SRS option is mandatory, it should not be hidden under "Advanced options". It is very unclear, why the "Add" button is disabled until the user opens the advanced options.
>    image

Moved outside the advanced options

DailisLangovskis:
> Some problems found during testing.
> 
> 1. I have added two images with nearly the same extent and nearly the same name. But both is different. I can see them in the list of my layers in the Datasources panel. But only one can be added to the map. When I try to add the other, it looks like it was added, but it's still the first one.
>    image
>    image
> 
> If I remove the jpg version and try to add the png only, it adds nothing. image
> 
> Update: After reloading the app, it is possible to add all the uploaded layers from datasources correctly 🤔 Albeit I have not seen any errors in the console before. image

Could It be that layman hasn't finished uploading the layer, but it is being added to the map?

DailisLangovskis:
> 2. There should be no option to add raster image as a WFS service.
>    image

I have no idea, how can we check it, because layman doesn't return any unique value for that.

fzadrazil:
> I have no idea, how can we check it, because layman doesn't return any unique value for that.

There is the file_type attribute for that in the GET Workspace Layer response
image
We are already sending this request when the "Add to map" is clicked, so we should be able to extract it from there.

fzadrazil:
> Well, maybe layer is not finished, but it is being added to the map, however i did fix this with geotiff previously.

Yes, it was probably the case when the layer was still being processed on Layman. We are, unfortunatelly, not able to simply get this info for all the layers from Layman at one time in the GET Workspace LayerS, but we could get it from GET Workspace Layer after the user picks one layer and clicks "Add to map" (similar to the previous check where we need to decide whather it is a vector or raster).

DailisLangovskis:
> > I have no idea, how can we check it, because layman doesn't return any unique value for that.
> 
> There is the file_type attribute for that in the GET Workspace Layer response image We are already sending this request when the "Add to map" is clicked, so we should be able to extract it from there.

Resolved

DailisLangovskis:
> > Well, maybe layer is not finished, but it is being added to the map, however i did fix this with geotiff previously.
> 
> Yes, it was probably the case when the layer was still being processed on Layman. We are, unfortunatelly, not able to simply get this info for all the layers from Layman at one time in the GET Workspace LayerS, but we could get it from GET Workspace Layer after the user picks one layer and clicks "Add to map" (similar to the previous check where we need to decide whather it is a vector or raster).

Going through the inspector, while loading new raster layer, it seems that the layer gets published, which is confirmed by `layman_metadata: {publication_status: 'COMPLETE'}`, but when the OWS service requests the get request from Layman, the newly added layer is not inside the service layers list yet.

DailisLangovskis:
Haven't resolved the loading issue, but seems that caching the response from layman on prev request, could be responsible, and also there might be some caching mechanism for Layman as well. Also noticable when deleting server Layman layers.

FilipLeitner:
Well its capabilities caching that makes typesrvice.listLayerFromCapabilities to fail - layer is not identified because cached capabilities doesnt include the layer. For firstly added layer it works because capabilities have not been cached yet.

So I guess we could just stop caching layman endpoin capabilities. Or add some boolean param  for capabilities `request` function that would prevent using cached capabilities  while updating it

DailisLangovskis:
> @DailisLangovskis Well its capabilities caching that makes typesrvice.listLayerFromCapabilities to fail - layer is not identified because cached capabilities doesnt include the layer. For firstly added layer it works because capabilities have not been cached yet.
> 
> So I guess we could just stop caching layman endpoin capabilities. Or add some boolean param for capabilities `request` function that would prevent using cached capabilities while updating under specified conditions. What do you think?

I think the cache should remain, otherwise external services, which sometimes need to be reloaded, could be annoyingly slow. It is most likely that a clear cache mechanism should be created for such cases.

fzadrazil (approved):

PR #2622, opened 2022-02-02
fix: Finding of correct layer when feature IDs overlap

### LINKED ISSUE #2617: Query info panel always displays top-most layer title
# Bug

**Describe the bug**
When a layer is queried, Info panel displays top-most layer title instead of active layer title. The properties displayed are correct.

**To Reproduce**
1. Add two overlapping layers
2. Hide one of them
3. Click on some feature in the active layer
4. See incorrect title (if this layer was not top-most, otherwise the title is correct)

**Expected behavior**
Correct title shall be displayed.

**Screenshots**
obrazek
obrazek

**Additional context**
Tested under 7.0.4

raitisbe:
## Description

Multiple layers can have features with the same id. In that case we need to loop through feature
collections and compare by feature instance instead.

fix #2617

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [x] Documentation content changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/layout/panels/panel-container.component.ts
+++ b/projects/hslayers/src/components/layout/panels/panel-container.component.ts
@@ -30,6 +30,10 @@ export class HsPanelContainerComponent implements OnInit, OnDestroy {
   /** Miscellaneous data object to set to each of the panels inside this container.
    * This is used if undefined value is passed to the create functions data parameter. */
   @Input() data: any;
+  /**
+   * Set this to true to not clear the ReplaySubject on container destruction because
+   * panels are added to ReplaySubject from app component and we cant re-add them. */
+  @Input() reusePanelObserver?: boolean;
   @Input() panelObserver?: ReplaySubject<HsPanelItem>;
   @Output() init = new EventEmitter<void>();
   interval: any;
@@ -39,7 +43,7 @@ export class HsPanelContainerComponent implements OnInit, OnDestroy {
     private HsConfig: HsConfig
   ) {}
   ngOnDestroy(): void {
-    if (this.service.panelObserver) {
+    if (this.service.panelObserver && this.reusePanelObserver !== true) {
       this.service.panelObserver.complete();
       this.service.panelObserver = new ReplaySubject<HsPanelItem>();
     }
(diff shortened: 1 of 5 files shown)


jmacura:
There is something weird happening.

I have upgraded the problematic application to this branch.

When I load the app and hover the mouse over the first feature, it works as it should.
obrazek

But when I move to some other feature (or leave the first one and then get back), it immediately breaks and the pop-up shows nothing. Console is clean however.
obrazek


The query info-panel works better. Only I would expect it to list all features beneath the point of query. Now it only displays the first one.
obrazek

While also others are turned on.
obrazek

But the initial problem with layer title is correct now! :-)

raitisbe:
>The query info-panel works better. Only I would expect it to list all features beneath the point of query. Now it only displays the first one.
Im not agains, but maybe changing that logic should be left for some other time and PR. In that case we should not use Select interaction at all but source.getFeaturesAtCoordinate() on each vector source and it would require quite a lot of rewriting.

raitisbe:
>But when I move to some other feature (or leave the first one and then get back), it immediately breaks and the pop-up shows
Will look on this. I did mostly clicking and my hover popup config is heavily modified so didnt notice.

jmacura:
> > The query info-panel works better. Only I would expect it to list all features beneath the point of query. Now it only displays the first one.
> 
> Im not against, but maybe changing that logic should be left for some other time and PR. In that case we should not use Select interaction at all but source.getFeaturesAtCoordinate() on each vector source and it would require quite a lot of rewriting.

OK, if this is intended behaviour, then let's leave it as it is now. I agree completely.

raitisbe:
Should be fixed. Noticed some minor flickering when entering previously not visited features, but it's better than it used to be.

jmacura:
obrazek
Fix confirmed!

jmacura (approved):
Good to go from my POV. Thank you!

raitisbe:
Flickering of phover popup is also solved in https://github.com/hslayers/hslayers-ng/commit/7e1432ae5e23702bc8bbab3cd40ef79e6aa4e589 . Was due to missing `return`

PR #2644, opened 2022-02-04
Layer scale settings changes in LayerManager when layers are switched

### LINKED ISSUE #2629: Layer scale settings in LayerManager keeps its value when layer switches
# Bug

**Describe the bug**
The scale property textbox keeps its value when layer settings in LayerManager are switched. More specifically it only keeps the value when you swich the settings of some other layer directly, without closing settings of the first one. When a visibility is turned on/off, the value in LayerManager resets to what was set for that layer originally.

**To Reproduce**
1. In LayerManager set scale for some layer
2. Switch settings of some _other layer_
4. See how the scale changes also for the _other layer_

**Expected behavior**
Layer settings shall be independent no matter how you open the settings

**Screenshots**
Rewrite to 100
obrazek
Close settings, open other layer settings
obrazek
Don't close settings, open other layer settings directly
obrazek

**Additional context**
At first I thought it's because the layers are copies created via 'Copy layer' but it does not seem to be that case.

DailisLangovskis:
## Description

changeDetection: ChangeDetectionStrategy.OnPush was causing the scale value to stay the same even when layers where changed without closing the scale widged

closes #2629 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/layermanager/widgets/scale-widget.component.ts
+++ b/projects/hslayers/src/components/layermanager/widgets/scale-widget.component.ts
@@ -1,4 +1,4 @@
-import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {Component} from '@angular/core';
 
 import {HsLayerEditorWidgetBaseComponent} from './layer-editor-widget-base.component';
 import {HsLayerSelectorService} from '../editor/layer-selector.service';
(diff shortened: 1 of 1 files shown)


jmacura (approved):
From my POV the issue is fixed, so I approve.

I am only wondering, whether the changeDetection strategy wasn't set for some reason. We usually don't do that just because. It would be unpleasant to fix one issue and create another.

DailisLangovskis:
> From my POV the issue is fixed, so I approve.
> 
> I am only wondering, whether the changeDetection strategy wasn't set for some reason. We usually don't do that just because. It would be unpleasant to fix one issue and create another.

As far as I remember, it was for optimization purposes.

PR #2649, opened 2022-02-07
Map swipe state in url, swipe control properties stored to local storage (Issue 2633)

### LINKED ISSUE #2633: Swipe state (enabled/disabled) is lost after page reload
# Feature Request

Swipe state (enabled/disabled) is lost after page reload. For convenience, its state shall be stored in URL params or in the sessionStorage.
The use case in mind is a map sharing: To be able to share a map which has the map-swipe panel enabled, but the map-swipe function currently turned off.

Add URL param like `swipe=on` (for brevity) or `map-swipe=enabled` (if verbosity is desired).

LocalStorage might be a way to save also other swipe-properties like swipe-side, swipe-right-layers, swipe-position, ...? But would be probably overkill for a URL.

**Additional context**
.

### LINKED ISSUE #2655: External vector layers won't load after page reload
# Bug

**Describe the bug**
When some vector layers from URL are added to the map, they are saved to localStorage when the page reloads, but they can't be loaded.

**To Reproduce**
1. Add some vector layer from URL, e.g. https://raw.githubusercontent.com/jmacura/sieusoil-website-map/main/data/Sieusoil_Pilots.geojson
2. Reload the page
3. See the layer in LayerManager, but not its content

**Expected behavior**
The layer should load

**Screenshots**
Before reload
obrazek

After reload
obrazek

**Additional context**
Looking into the localStorage, the layer is there including its URL but HSL won't even try to load it (no request present in the Network tab of dev tools).
```
{
      "metadata":{
        
      },
      "visibility":true,
      "swipeSide":"right",
      "opacity":1,
      "base":false,
      "title":"sieusoil",
      "path":"other",
      "name":"sieusoil",
      "className":"Vector",
      "protocol":{
        "url":"https%3A%2F%2Fraw.githubusercontent.com%2Fjmacura%2Fsieusoil-website-map%2Fmain%2Fdata%2FSieusoil_Pilots.geojson",
        "format":"hs.format.WFS"
      },
      "workspace":"macurajan",
      "maxResolution":null,
      "minResolution":0,
      "projection":"epsg:4326",
      "style":"<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>\n<StyledLayerDescriptor version=\"1.0.0\" xsi:schemaLocation=\"http://www.opengis.net/sld StyledLayerDescriptor.xsd\" xmlns=\"http://www.opengis.net/sld\" xmlns:ogc=\"http://www.opengis.net/ogc\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">\n  <NamedLayer>\n    <Name/>\n    <UserStyle>\n      <Name/>\n      <Title/>\n      <FeatureTypeStyle>\n        <Rule>\n          <Name/>\n          <PolygonSymbolizer/>\n        </Rule>\n      </FeatureTypeStyle>\n    </UserStyle>\n  </NamedLayer>\n</StyledLayerDescriptor>"
    }
```

DailisLangovskis:
## Description

This PR provides requested features from the feature issue, such as, map-swipe custom URL parameter, that is used to check the state of the swipe control state.

Map swipe control position and orientation is now stored to local storage.
When the map composition is being shared or stored to local storage, swipeSide property is set for each layer, so it can be obtained.

closes #2633 
closes #2655 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/add-data/vector/vector-layer-options.type.ts
+++ b/projects/hslayers/src/components/add-data/vector/vector-layer-options.type.ts
@@ -15,4 +15,5 @@ export type HsVectorLayerOptions = {
   access_rights?: accessRightsModel;
   queryCapabilities?: boolean;
   sld?: string;
+  saveToLayman?: boolean;
 };
--- a/projects/hslayers/src/components/add-data/vector/vector.service.ts
+++ b/projects/hslayers/src/components/add-data/vector/vector.service.ts
@@ -97,16 +97,19 @@ export class HsAddDataVectorService {
         TODO: Should have set definition property with protocol inside 
         so layer synchronizer would know if to sync 
         */
-        if (this.hsUtilsService.undefineEmptyString(url) !== undefined) {
-          setDefinition(lyr, {
-            format: 'hs.format.WFS',
-            url: url,
-          });
-        } else {
-          setDefinition(lyr, {
-            format: 'hs.format.WFS',
-          });
+        if (options.saveToLayman) {
+          if (this.hsUtilsService.undefineEmptyString(url) !== undefined) {
+            setDefinition(lyr, {
+              format: 'hs.format.WFS',
+              url: url,
+            });
+          } else {
+            setDefinition(lyr, {
+              format: 'hs.format.WFS',
+            });
+          }
         }
+
         if (this.hsMapService.map) {
           this.hsAddDataService.addLayer(lyr, addUnder);
         }
@@ -177,12 +180,10 @@ export class HsAddDataVectorService {
         ? new sourceDescriptor.sourceClass(sourceDescriptor.sourceParams)
         : new sourceDescriptor.sourceClass(sourceDescriptor);
     descriptor.layerParams.source = src;
-    if (descriptor.layerParams.style) {
-      Object.assign(
-        descriptor.layerParams,
-        await this.hsStylerService.parseStyle(descriptor.layerParams.style)
-      );
-    }
+    Object.assign(
+      descriptor.layerParams,
+      await this.hsStylerService.parseStyle(descriptor.layerParams.style)
+    );
     const lyr = new VectorLayer(descriptor.layerParams);
     return lyr;
   }
@@ -239,6 +240,7 @@ export class HsAddDataVectorService {
           data.dataType != 'kml' &&
           data.dataType != 'gpx' &&
           !data.url?.endsWith('json'),
+        saveToLayman: data.saveToLayman,
       },
       data.addUnder
     );
--- a/projects/hslayers/src/components/compositions/compositions-parser.service.ts
+++ b/projects/hslayers/src/components/compositions/compositions-parser.service.ts
@@ -25,6 +25,7 @@ import {
   getFromComposition,
   getTitle,
   setMetadata,
+  setSwipeSide,
 } from '../../common/layer-extensions';
 import {parseExtent, transformExtentValue} from '../../common/extent-utils';
 
@@ -477,6 +478,7 @@ export class HsCompositionsParserService {
     }
     if (resultLayer) {
       setMetadata(resultLayer, lyr_def.metadata);
+      setSwipeSide(resultLayer, lyr_def.swipeSide);
     }
     return resultLayer;
   }
--- a/projects/hslayers/src/components/map-swipe/map-swipe.service.mock.ts
+++ b/projects/hslayers/src/components/map-swipe/map-swipe.service.mock.ts
@@ -1,6 +1,10 @@
 export function mockHsMapSwipeService() {
   return jasmine.createSpyObj('HsMapSwipeService', [
-    'init',
+    'setInitCtrlActive',
+    'setInitOri',
+    'updateUrlParam',
+    'updateStorageOri',
+    'initSwipeControl',
     'layersAvailable',
     'setOrientation',
     'fillSwipeLayers',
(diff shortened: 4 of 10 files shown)


jmacura (commented):
The swipe enabling/disabling via URL works fine. The swipe position is kept as well. 👍 

Do I understand it correctly that only the initial swipeSide property is stored in the localStorage?
Because when I re-arrange the layers between swipe sides, this information does not persist after page reload.

DailisLangovskis:
Well, if You tested this with test app default layers, then nothing is
stored, for added layers, that are not default ones, swipeSides are kept.
jmacura ***@***.***> (šajā datumā: otrd., 2022. g. 8. febr.
02:52) rakstīja:

raitisbe:
Please test if a composition can now be saved to layman since a new property swipeSide is set.

DailisLangovskis:
> Please test if a composition can now be saved to layman since a new property swipeSide is set.

I just tried it, saved successfully!

jmacura:
> Well, if You tested this with test app default layers (points, polygons, etc.), then stored information to local storage is not changed, at least that is what is happening. Other layers, will be stored with the swipeSide property as expected. P.S. swipeSide property is stored with the whole layer information, when cookies are being used.

OK, so it basically works only with the catalogue layers, right?

DailisLangovskis:
> > Well, if You tested this with test app default layers (points, polygons, etc.), then stored information to local storage is not changed, at least that is what is happening. Other layers, will be stored with the swipeSide property as expected. P.S. swipeSide property is stored with the whole layer information, when cookies are being used.
> 
> OK, so it basically works only with the catalogue layers, right?

Yes, any layer that is not predefined. The only question, is such behavior planned that changes to predefined layers are not taken into account?

DailisLangovskis:
> I tested succesfully with WMS layer obrazek
> 
> But my test with a GeoJSON layer was unsuccesful due to #2655 .
> 
> > The only question, is such behavior planned that changes to predefined layers are not taken into account?
> 
> I can see that pre-defined layers are stored in the localStorage as well. So is there any reason why not to save swipeSide for them too?

Well I have set the swipeSide to the layer, but why it is not stored inside the storage, i don't know, mby I have missed something more @raitisbe ?

DailisLangovskis:
> > I tested succesfully with WMS layer obrazek
> > But my test with a GeoJSON layer was unsuccesful due to #2655 .
> > > The only question, is such behavior planned that changes to predefined layers are not taken into account?
> > 
> > 
> > I can see that pre-defined layers are stored in the localStorage as well. So is there any reason why not to save swipeSide for them too?
> 
> Well I have set the swipeSide to the layer inside layer2json, but why it is not stored for predefined layers inside the storage, i don't know, mby I have missed something more @raitisbe ?

jmacura:
When I am on this branch and do start-test-app, I also receive this lengthy error in the console
obrazek
Few lines lower it ends with.
obrazek

I can't experience that when on develop branch..

DailisLangovskis:
> When I am on this branch and do start-test-app, I also receive this lengthy error in the console obrazek Few lines lower it ends with. obrazek
> 
> I can't experience that when on develop branch..

Working on it as well, because the style is not correctly parsed and stored, if it was set as createDefaultStyle function.

jmacura:
> Well, the reason for this is here -> https://github.com/hslayers/hslayers-ng/blob/develop/projects/hslayers/src/components/compositions/compositions.service.ts#L287, if we would change the ignoreNew to RemoveOriginal, it would work for predefined as well. Question is, do we want it to be like so @raitisbe @jmacura ?

Good point. I was searching through the JSON file saved inside _hs_layers_ localStorage and the `swipeSide` values are indeed there.
Excerpt:
```
{
      "metadata":{  
      },
      "visibility":true,
      "swipeSide":"right",
      "opacity":1,
      "base":false,
      "title":"Polygons",
      "path":"User generated",
      "name":"Polygons",
      "className":"Vector",
...
```

So... I do not know what was the ratio behind IgnoreNew strategy there. Not sure if the RemoveOriginal would have some negative consequences.

projects/hslayers/src/components/compositions/compositions.service.ts
```diff
@@ -284,7 +284,7 @@ export class HsCompositionsService {
         parsed
       );
       for (let i = 0; i < layers.length; i++) {
-        this.hsMapService.addLayer(layers[i], DuplicateHandling.IgnoreNew);
+        this.hsMapService.addLayer(layers[i], DuplicateHandling.RemoveOriginal);
```
DailisLangovskis, line 287: If IgnoreNew any layer with created faulty style, bad name, missing features, after fixing and refreshing, will be ignored anyway.

raitisbe: Time based and layer configuredd inside app.component.ts probalby were the reason for using IgnoreNew indeed.

DailisLangovskis: @raitisbe What should be one about this then?

jmacura:
FTR, the implemented behaviour is such that it remembers the last swipeSide for layers which are not default_layers. Correct?

Before page reload
obrazek

After page reload
obrazek

DailisLangovskis:
Well yes, because of the ignoreNew handler in compositions-parser service

ceturtd., 2022. g. 10. febr., plkst. 15:37 — lietotājs jmacura (<
***@***.***>) rakstīja:

jmacura:
> Well yes, because of the ignoreNew handler in compositions-parser service 

OK, fine with me. Thanks for all the changes!

PR #2794, opened 2022-03-18
Ability to set and add IDW - Interpolated layer to Hslayers (issue-2747)

### LINKED ISSUE #2747: IDW Layer
# Feature Request

Implement IDW Layer, more details at http://viglino.github.io/ol-ext/examples/layer/map.layer.idw.html

DailisLangovskis:
## Description

Interpolated layers can be added to Hslayers using InterpolatedSource class.

```
const interpolatedSource = new InterpolatedSource({
        maxFeaturesInCache: 500,
        maxFeaturesInExtent: 100,
        features: [],
        weight: 'fac2020',
        loader: async ({extent, projection}) => {
          if (this.timer !== null) {
            clearTimeout(this.timer);
          }
          interpolatedSource.cancelUrlRequest.next();
          const extentIn4326 = transformExtent(extent, projection, 'EPSG:4326');
          const url = this.hsUtilsService.proxify(
            interpolatedSource.createIDWSourceUrl(
              'https://api-agroclimatic.lesprojekt.cz/area/selection/preci/0/{minY}/{maxY}/{minX}/{maxX}/100/random/year/2020/2020/1/5/2020-01-01/2020-01-30/1/1/ERA5-Land',
              extentIn4326
            ),
            app.name
          );
          try {
            const response: any = await lastValueFrom(
              this.httpClient.get(url).pipe(
                takeUntil(interpolatedSource.cancelUrlRequest),
                catchError(async (e) => {})
              )
            );
            if (response?.features?.length > 0) {
              return interpolatedSource.parseFeatures(response, projection);
            }
          } catch (error) {}
        },
        colorMap: 'jet',
      });
```

Then the layer can be created:

```
  const idwLayer = new ImageLayer({
        properties: {title: 'IDW layer'},
        source: interpolatedSource as any,
        opacity: 0.5,
      });
```

In case when a custom `loader` function is provided to load features from external resource a separate Vector layer needs to be created. The underlying vector source is linked to that (`interpolatedSource.getSource()`), otherwise the feature `loader` function won't be called. It can also be used to display the values being interpolated - both normalized and original ones. Provide a `Style` to display labels in that case. Set initial visibility from interpoloated ImageLayer, so it is synchronized.

```
      //Mandatory, otherwise nothing will be loaded with source loader
      const idwVectorLayer = new VectorLayer({
        properties: {
          showInLayerManager: false,
          visible: idwLayer.getVisible(),
        },
        style: new Style(),
        source: interpolatedSource.getSource(),
      });
```

Provide interpolated layer visibility change listener and update vector layer visibility, based on the changes. This will disable feature loading from external service, if the interpolated layer is invisible.

```
idwLayer.on('change:visible', (e) => {
        idwVectorLayer.setVisible(e.target.getVisible());
      });
```

The layer is drawn using 'jet' color map by default 
image
. To add other custom color maps ->

1. npm i colormap @types/colormap
2. Add import createColormap from 'colormap';
3. Get desired colormap with ->
```
const colors = createColormap({
      colormap: 'jet',
      nshades: 100,
      format: 'rgba',
      alpha: 1,
    });
```
4. Set colorMap function in InterpolatedSource constructor options as you desire.
Example: 
```
colorMap: (v) => {
          let color = [0, 0, 0, 255];
          if (isNaN(v)) {
            return color;
          }
          v = Math.floor(v);
          switch (true) {
            case v < 20:
              color = colors[10];
              break;
            case v < 40:
              color = colors[30];
              break;
            case v < 60:
              color = colors[50];
              break;
            case v < 80:
              color = colors[70];
              break;
            case v <= 0:
            case v > 99:
              color = colors[90];
              break;
            default:
              color = [0, 0, 0, 255];
          }
          return color;
        },
```

closes #2747 

## Pull request type

- [x] Bugfix
- [x] Feature
- [x] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist


DailisLangovskis:
This is my proposal of how to implement this layer using config. Still missing updates on extent and bounding_box check

raitisbe:
I think we should consider IDW layers to be similar to sparqlLayer or Wfs and declare them in HsConfig.default_layers to be consistent. There could be multiple interpolated layers, so a new layer class needs to be created which describe those. This way it will be easier to save/load them to compositions display in layer manager etc.

projects/hslayers/src/common/layers/public-api.ts
```diff
@@ -1,3 +1,5 @@
 export * from './hs.source.SPOI';
 export * from './hs.source.SparqlJson';
 export * from './hs.source.WfsSource';
+export * from './interpolated-layer.class';
```
jmacura, line 4: Don't name the plain class a ".class". Everything is a class. Let's go with just "interpolated-layer.ts".

DailisLangovskis: Sorry about that!

projects/hslayers/src/common/layers/interpolated-source.ts
```diff
@@ -0,0 +1,190 @@
...
+  };
+}
+
+export class InterpolatedSource extends IDW {
+  cancelUrlRequest: Subject<void> = new Subject();
+  constructor(private options: InterpolatedSourceOptions) {
+    super({
+      // Source that contains the data
+      source: new VectorSource({
+        features: options.features,
```
raitisbe, line 38: Need to normalize those too

DailisLangovskis: Do we want to keep those initial features all the time?

raitisbe: no, just until loader is called if it's provided

projects/test-app/src/hslayers-app/hslayers-app.component.ts
```diff
@@ -37,18 +44,82 @@ export class HslayersAppComponent {
...
+          }
+          return color;
+        },
+      });
+      const idwLayer = new ImageLayer({
+        properties: {title: 'IDW layer'},
+        source: interpolatedSource as any,
+        opacity: 0.5,
+      });
+      //Manditory, otherwise nothing will be loaded with source loader
```
raitisbe, line 121: But it is strange, since the loader is set on interpolatedSource. If its mandatory, it should be created from inside interpolatedSource class, but I still have my doubts

projects/hslayers/src/common/layers/public-api.ts
```diff
@@ -1,3 +1,4 @@
 export * from './hs.source.SPOI';
 export * from './hs.source.SparqlJson';
 export * from './hs.source.WfsSource';
+export * from './interpolated-source';
```
jmacura, line 4: I am not pushing you to anything, but I would name the file `hs.source.interpolated` just for the sake of consistency in this folder.

projects/hslayers/src/components/layermanager/layermanager.service.ts
```diff
@@ -365,6 +365,10 @@ export class HsLayerManagerService {
     if (this.HsLayerUtilsService.isLayerArcgis(layer)) {
       return 'ARCGIS';
```
jmacura, line 366: btw, this should be `ArcGIS`, but.. a nitpick indeed 🙂

projects/hslayers/src/components/layermanager/layermanager.service.ts
```diff
@@ -365,6 +365,10 @@ export class HsLayerManagerService {
...
     }
+
+    if (this.HsLayerUtilsService.isLayerIDW(layer)) {
+      return 'IDW';
```
jmacura, line 370: I am not sure now about the actual implementation.. Is the layer always interpolated via the IDW algorithm? If so, then the return 'IDW' is perfectly fine, if other interpolation methods could be used then it would be misleading.

projects/hslayers/src/common/layers/interpolated-source.ts
```diff
@@ -0,0 +1,200 @@
...
+                extent,
+                resolution,
+                projection,
+                success,
+                failure,
+              }),
+              extent
+            );
+          }
+        },
```
jmacura, line 66: Initially I thought that we could provide a default loader like the one that is now used in the test-app example. By just adding `url` as yet another optional parameter to InterpolatedSourceOptions. But I don't know now.. seems to be too intertwined with other Angular-ish code, like proxying, cancelling requests etc.
So I am just leaving it here as a thought 🙃

projects/hslayers/src/common/layers/interpolated-source.ts
```diff
@@ -0,0 +1,200 @@
...
+      src.addFeatures(features);
+    }
+  }
+
+  /**
+   * Parse features from get request
+   * @param collection - Get request response feature collection
+   * @param mapProjection - Map projection
+   */
+  parseFeatures(collection: any, mapProjection: string | Projection): void {
```
jmacura, line 132: `collection: any` but then we rely on its `features` property. Not much safety in that...

projects/hslayers/src/common/layers/interpolated-source.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+    return url;
+  }
+  /**
+   * Normalize weight values to be between 0 and 100
+   * @param features - OL feature array
+   * @param weight - Weight property name
+   */
+  normalizeWeight(weight: string): void {
+    //https://www.statology.org/normalize-data-between-0-and-100/
```
jmacura, line 184: why not moving this line into the tsdoc comment above?

projects/hslayers/src/components/map/map.service.ts
```diff
@@ -85,21 +82,20 @@ export class HsMapService {
...
-  visibleLayersInUrl;
+  visibleLayersInUrl: string[];
   //timer variable for extent change event
   timer = null;
   puremap: any;
   /**
    * @public
-   * @type {number} 400
-   * @description Duration of added interactions animation. (400 ms used, default in OpenLayers is 250 ms)
+   * 400
```
jmacura, line 91:
```suggestion
   * @default 400
```

projects/hslayers/src/components/map/map.service.ts
```diff
@@ -291,22 +309,29 @@ export class HsMapService {
...
   }
 
-  setDefaultView = function (e, app) {
+  /**
+   * @public
+   * Set map to default view
+   * @param e - Mouse click event
+   * @param app - App identifier
+   */
+  setDefaultView = function (e, app): void {
```
jmacura, line 318: ❤️ for all these doc comments in this service!

projects/hslayers/src/components/map/map.service.ts
```diff
@@ -602,11 +637,12 @@ export class HsMapService {
   }
 
   /**
-   * @description Checks if a layer with the same title already exists in the map
-   * @param {ol/Layer} lyr A layer to check
-   * @returns {boolean} True if layer is already present in the map, false otherwise
+   * Checks if a layer with the same title already exists in the map
+   * @param lyr - A layer to check
+   * @param app - App identifier
+   * @returns True or false
```
jmacura, line 643:
```suggestion
   * @returns True if layer is already present in the map, false otherwise
```

projects/hslayers/src/components/map/map.service.ts
```diff
@@ -927,22 +1005,35 @@ export class HsMapService {
...
     }
   }
 
-  laymanWmsLoadingFunction(image, src: string) {
+  /**
+   * @public
+   * @param image -
+   * @param src -
+   *
+   */
```
jmacura, line 1035: 💔 but these empty comments are just useless lines of code..

jmacura (commented):
https://bundlephobia.com/package/ol-ext@3.2.22 ol-ext is adding ~650kB to the bundle. Not great not terrible.

projects/hslayers/src/components/layermanager/layermanager.service.ts
DailisLangovskis: Currently only IDW, as far as I know.

projects/hslayers/src/components/map/map.service.ts
DailisLangovskis: I know, but I was not sure about what to write there, would appreciate some additional descriptions. :)

projects/hslayers/src/common/layers/interpolated-source.ts
DailisLangovskis: No problem! :)

projects/hslayers/src/common/layers/interpolated-source.ts
DailisLangovskis: Can You suggest any better type for it?

raitisbe:
> https://bundlephobia.com/package/ol-ext@3.2.22 ol-ext is adding ~650kB to the bundle. Not great not terrible.

It is actually much less. 650 is to he nodemodules directory but olext/source is about 2kb gziped. Checked with webpack bundleanalyzer

projects/hslayers/src/common/layers/interpolated-source.ts
raitisbe: The reason why its like this is the custom format how the features might be returned from the server. Hypothetically it could even be some csv file which doesn't provide feature geometries, but just IDs and values.

jmacura: I see.

projects/hslayers/src/components/layermanager/layermanager.service.ts
jmacura: Acknowledged.

projects/hslayers/src/common/layers/interpolated-source.ts
jmacura: From the usage inside the method, it would be obviously
```
{
  crs?: string;
  srs?: string;
  features: string[];
}
```

but given the actual use of this method, which we do not use internally but is only called from the loader in the test-app, I would say moving the if-condition from there to here might be sufficient (and keeping the 'any').
I.e. the line `if (response?.features?.length > 0) {` gives me much more sense inside here than in every custom loader a consumer of hslayers would need to write.

Does it make sense to you? 😄

projects/hslayers/src/components/map/map.service.ts
jmacura: Hmm.. I will take a look at the map.service and will try to add something.

fzadrazil (approved):
Works great from the user's point of view.

projects/hslayers/src/components/map/map.service.ts
jmacura: A commit is there 😉

jmacura (approved):
> Works great from the user's point of view.

And that is by far the most important thing. If the implementation runs smoothly, then this PR is good to go.

Issue #2836, opened 2022-04-01
Support CSW type of map composition


fzadrazil:
# Feature Request

**Is your feature request related to a problem? Please describe.**
There is a type of map composition in Micka, which does not have other definition than CSW metadata record itself. We need to load and display such compositions in the Compositions panel.

**Describe the solution you'd like**
There is a new JSON format Micka can return the records in. The only difference compared to the currently used request is the added url parameter `template=report-layman`. The request returns identical JSON data as up until now, but there is a new element operatesOn, which contains links to the referenced map services. Those services should be loaded in the map.
The CSW composition type can be distinguished by the parameter serviceType=CSW in the json response.

**Additional context**
Sample request (there is one CSW composition in the response):
https://hub.lesprojekt.cz/micka/csw/?request=GetRecords&query=type%3D%27application%27%20AND%20AnyText%20like%20%27*pozemko*%27&format=text/json&MaxRecords=10&StartPosition=&sortby=&language=eng&template=report-layman

DailisLangovskis:
I tried adding -> `&template=report-layman` to Micka request url in https://github.com/hslayers/hslayers-ng/blob/develop/projects/hslayers/src/components/compositions/endpoints/compositions-micka.service.ts#L84, and I got this response ->
Records + 
```

<meta charset="utf-8">
<meta name="robots" content="noindex">
<title>Server Error</title>

<style>
	#error-body { background: white; width: 500px; margin: 70px auto; padding: 10px 20px }
	#error-body h1 { font: bold 47px/1.5 sans-serif; background: none; color: #333; margin: .6em 0 }
	#error-body p { font: 21px/1.5 Georgia,serif; background: none; color: #333; margin: 1.5em 0 }
	#error-body small { font-size: 70%; color: gray }
</style>

<div id="error-body">
    <h1>Server Error</h1>
    <p>We're sorry! The server encountered an internal error and
    was unable to complete your request. Please try again later.</p>
    <p><small>error 500</small></p>
</div>
```

FilipLeitner:
Yup, a bit more need to be changed. 
![image](https://user-images.githubusercontent.com/44566616/161528575-16daffba-c358-4b67-9227-e2e00d15f8b6.png)

PR #2852, opened 2022-04-06
Csw compositions

FilipLeitner:
## Description

Allow CSW compostions 

#2836 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/add-data/url/add-data-ows.service.ts
+++ b/projects/hslayers/src/components/add-data/url/add-data-ows.service.ts
@@ -158,7 +158,7 @@ export class HsAddDataOwsService {
 
   async setTypeServices(app: string): Promise<void> {
     const appRef = this.get(app);
-    switch (this.hsAddDataUrlService.apps[app].typeSelected) {
+    switch (this.hsAddDataUrlService.get(app).typeSelected) {
       case 'wmts':
         appRef.typeService = this.hsUrlWmtsService;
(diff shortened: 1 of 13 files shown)


fzadrazil:
It would probably be better to load a WMS service as one instead of each layer separatelly. That is a general issue, but can be really painfull with those csw compositions. Eg. https://gis.cenia.cz/geoserver/corine_land_cover_2018/wms has four data layers and one root "abstract" layer without a name. But it could still be added as a group and requested via a single GetMap request.

FilipLeitner:
> It would probably be better to load a WMS service as one instead of each layer separatelly. That is a general issue, but can be really painfull with those csw compositions. Eg. https://gis.cenia.cz/geoserver/corine_land_cover_2018/wms has four data layers and one root "abstract" layer without a name. But it could still be added as a group and requested via a single GetMap request.

By group you mean sublayers? Because I dont think its that much of a good idea to merge all layers ( first queriable level in layer tree which we are loading now) into sigle layer without a possiblity to control them separately because

1.  if there are more layers covering same area  we wolud be able to display only one 
2. in case the layer order is not correct we would also lose some features etc.

fzadrazil (approved):
Basic version is fine like this. Loading services will be further improved #2873

PR #2916, opened 2022-04-19
fix: updSidebarpoisiton not emmiting values

FilipLeitner:
updSidebarpoisiton  was not getting subject value using `lastValueFrom` thus not setting sidebarPosition value for comonents subscribed. Resulting in for example  overflowing sidebar when collapsed - see: https://hub4everybody.lesprojekt.cz/land-use/


raitisbe:
Thank you

PR #2935, opened 2022-04-21
refactor: Allow dynamic weight parameter for interpolated layer

FilipLeitner:


DailisLangovskis:
I assume You have tested it, so I have no problem with the change. Only what will be the use case for it?

FilipLeitner:
> I assume You have tested it, so I have no problem with the change. Only what will be the use case for it?

https://api-agroclimatic.lesprojekt.cz/area/selection returns values in format such as fac[selectedyear] so the weight attr changes

DailisLangovskis (approved):

Issue #2984, opened 2022-05-05
Reduce function calling from templates on each digest loop.


raitisbe:
# Bug

Currently in PR https://github.com/hslayers/hslayers-ng/pull/2983 77 places have been marked in templates where function is called using `<!-- TODO: Remove function call from template -->`
**Expected behavior**
Some solutions how to refactor this are described in
https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496

PR #2988, opened 2022-05-09
perf: Reduce function calls in templates

raitisbe:
## Description
This PR reduces function calls from templates.

+ Provide isVisible$ BehaviourSubject in panel components for which the value is set when user sets current panel either by sidebar button or programmatically using setMainPanel function. In template `[hidden]="(isVisible$ | async) === false"` must be used instead of `[hidden]="isVisible()"`.
+ Layer titles are translated using an improved translateHs pipe even though the performance benefits are negligible since our translateHs pipe as well as original translate from ngx-translate is impure thus also is called on each digest loop - same as a function.
+ AssetPath reading is removed from HsUtilsService and it is set inside `HsConfig.update` function instead. Also loading animation `ajaxLoaderPath` is set there

[(https://github.com/hslayers/hslayers-ng/issues/2984)](https://github.com/hslayers/hslayers-ng/issues/2984)

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [x] Other (please describe)

- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/projects/hslayers-sensors/src/components/sensors/partials/sensors.component.html
+++ b/projects/hslayers-sensors/src/components/sensors/partials/sensors.component.html
@@ -1,4 +1,4 @@
-<div class="card panel-default hs-main-panel" [hidden]="!isVisible()"><!-- TODO: Remove function call from template -->
+<div class="card panel-default hs-main-panel" [hidden]="(isVisible$ | async) === false">
     <hs-panel-header name="sensors" [app]="data.app" [title]="'PANEL_HEADER.SENSORS' | translateHs: data">
     </hs-panel-header>
     <div class="card-body">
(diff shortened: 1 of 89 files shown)


DailisLangovskis:
@raitisbe Getting error for HsQueryPopupComponent not having this new isVisible$ BehaviorSubject property, other than that, rest of it seems ok.

raitisbe:
> @raitisbe Getting error -> componentRefInstance.isVisible is not a function. Issue comes from HsQueryPopupComponent, because it implements HsDialogComponent, which has no isVisible$ BehaviourSubject or isVisible() function

Should be alright now. true will be used if isVisible is missing.

PR #3366, opened 2022-09-18
Initial layer loading refactoring

### LINKED ISSUE #3335: Initial layer loading
At the initiation of an app, we try to load a layer from 3/4 different sources:

1. box_layers and  default_layers from app config 
https://github.com/hslayers/hslayers-ng/blob/3e9aabb47d428b8d6d6526fd2ddf038c7dcc2958/projects/hslayers/src/components/map/map.service.ts#L517

2.  hs-composition URL parameter or defaultComposition config param
https://github.com/hslayers/hslayers-ng/blob/6a819463c0f17a72ad6937c4f84b8ae2abc782a8/projects/hslayers/src/components/compositions/compositions.service.ts#L87

3. Permalink layers
https://github.com/hslayers/hslayers-ng/blob/6a819463c0f17a72ad6937c4f84b8ae2abc782a8/projects/hslayers/src/components/compositions/compositions.service.ts#L88

Both `tryParseCompositionFromUrlParam `and `parsePermalinkLayers `remove all the layers satisfying this condition
https://github.com/hslayers/hslayers-ng/blob/3e9aabb47d428b8d6d6526fd2ddf038c7dcc2958/projects/hslayers/src/components/map/map.service.ts#L1173-L1177

Many times it might not be necessary to try loading all the layers from 1....at the same time layers from 2. should not be loaded in case permalink is being opened at all.
It would be really beneficial to  clean this up a little bit, imo.

We could probably filter box and default layers by removable param in case permalink or composition is supposed to be loaded.

Feels like a refactor which could have some unexpected results or culprits I am not aware of atm . What do you think?

DailisLangovskis:
## Description

Filter default_layers and box_layers by removable property if default composition or hs-composition in URL is set.

Do not add any hsConfig's layers, if hs-permalink in URL is set.

Set visibleLayersInUrl for each app's map service.

closes #3335 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/compositions/compositions.service.ts
+++ b/projects/hslayers/src/components/compositions/compositions.service.ts
@@ -85,7 +85,7 @@ export class HsCompositionsService {
       ? this.parsePermalinkLayers(permalink, _app)
       : this.tryParseCompositionFromUrlParam(_app);
 
-    if (this.hsConfig.get(_app).saveMapStateOnReload) {
+    if (this.hsConfig.get(_app).saveMapStateOnReload && !permalink) {
       //Load composition data from cookies only if it is anticipated
       setTimeout(() => {
         this.tryParseCompositionFromCookie(_app);
--- a/projects/hslayers/src/components/compositions/compositions.service.ts
+++ b/projects/hslayers/src/components/compositions/compositions.service.ts
@@ -80,8 +80,11 @@ export class HsCompositionsService {
    * @param _app - App identifier
    */
   init(_app: string) {
-    this.tryParseCompositionFromUrlParam(_app);
-    this.parsePermalinkLayers(_app);
+    const permalink = this.HsShareUrlService.getParamValue(HS_PRMS.permalink);
+    permalink
+      ? this.parsePermalinkLayers(permalink, _app)
+      : this.tryParseCompositionFromUrlParam(_app);
+
     if (this.hsConfig.get(_app).saveMapStateOnReload) {
       //Load composition data from cookies only if it is anticipated
       setTimeout(() => {
@@ -354,12 +357,8 @@ export class HsCompositionsService {
    * Load layers received through permalink to map
    * @param app - App identifier
    */
-  async parsePermalinkLayers(app: string): Promise<void> {
+  async parsePermalinkLayers(permalink: string, app: string): Promise<void> {
     await this.hsMapService.loaded(app);
-    const permalink = this.HsShareUrlService.getParamValue(HS_PRMS.permalink);
-    if (!permalink) {
-      return;
-    }
     const layersUrl = this.hsUtilsService.proxify(permalink, app);
     const response: any = await lastValueFrom(this.http.get(layersUrl));
     if (response.success == true) {
--- a/projects/hslayers/src/components/compositions/compositions.service.ts
+++ b/projects/hslayers/src/components/compositions/compositions.service.ts
@@ -356,9 +356,7 @@ export class HsCompositionsService {
    */
   async parsePermalinkLayers(app: string): Promise<void> {
     await this.hsMapService.loaded(app);
-    const permalink = this.HsShareUrlService.getParamValue(
-      HS_PRMS.permalink
-    );
+    const permalink = this.HsShareUrlService.getParamValue(HS_PRMS.permalink);
     if (!permalink) {
       return;
     }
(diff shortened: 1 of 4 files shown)


projects/hslayers/src/components/map/map.service.ts
```diff
@@ -791,23 +793,50 @@ export class HsMapService {
...
       appConfig.box_layers.forEach((box) => {
-        for (const lyr of box.getLayers().getArray() as Layer<Source>[]) {
-          this.addLayer(
-            lyr,
-            app,
-            DuplicateHandling.IgnoreNew,
-            visibilityOverrides
-          );
-        }
+        boxLayers = boxLayers.concat(
```
FilipLeitner, line 798: Here, I'd like Array.map better. Seems more readable for this use case imo

projects/hslayers/src/components/map/map.service.ts
```diff
@@ -791,23 +793,50 @@ export class HsMapService {
...
+        );
       });
+      this.addLayersFromAppConfig(boxLayers, visibilityOverrides, app);
     }
 
     if (appConfig.default_layers) {
-      const layers = appConfig.default_layers.filter((lyr) => lyr);
-      if (layers.length > 0) {
+      const defaultLayers: Layer[] = appConfig.default_layers.filter(
+        (lyr) => lyr
```
FilipLeitner, line 809: Are you expecting a defaultLayers item to be falsy somehow? Other wise the filter would be not needed I think

projects/hslayers/src/components/compositions/compositions.service.ts
```diff
@@ -80,8 +80,11 @@ export class HsCompositionsService {
...
    */
   init(_app: string) {
-    this.tryParseCompositionFromUrlParam(_app);
-    this.parsePermalinkLayers(_app);
+    const permalink = this.HsShareUrlService.getParamValue(HS_PRMS.permalink);
+    permalink
+      ? this.parsePermalinkLayers(permalink, _app)
+      : this.tryParseCompositionFromUrlParam(_app);
+
     if (this.hsConfig.get(_app).saveMapStateOnReload) {
```
FilipLeitner, line 88: We should probably skip this when permalink is true, as well. What do you think?

FilipLeitner (changes requested):

projects/hslayers/src/components/map/map.service.ts
DailisLangovskis: This was not my addition, so i am not sure, so I kept it for now.

projects/hslayers/src/components/compositions/compositions.service.ts
DailisLangovskis: Well, this was added based on Your comment, if permalink is there, we don't want to add defaultComposition, for example.

projects/hslayers/src/components/map/map.service.ts
FilipLeitner: Oh, sorry. Just a safety measure 'what if'..I guess :D lets leave it than

projects/hslayers/src/components/map/map.service.ts
DailisLangovskis: Hmm.. I used concat, because each box is a group, and every group layers gets added to existing array of layers. I think map will only add last groups layers to boxLayers array.

projects/hslayers/src/components/compositions/compositions.service.ts
FilipLeitner: Yes sure thats all good. I just mean the next in condition with saveMapStateOnReload. ...that it would be better to skip it as well.
like: `if (this.hsConfig.get(_app).saveMapStateOnReload && !permalink)` 

I can see that inside `tryParseCompositionFromCookie` function there is some kind of condition for permalink as well 
https://github.com/hslayers/hslayers-ng/blob/34cf87ad7cf6ced36feaa9f546a109f8fc0baac7/projects/hslayers/src/components/compositions/compositions.service.ts#L411-L412

but don't see where its being set. So its not of much use as of right now

projects/hslayers/src/components/map/map.service.ts
FilipLeitner: I see it now. My bad, sorry

FilipLeitner (dismissed):

FilipLeitner (approved):

Issue #3515, opened 2022-11-05
Some ArcGIS services won't load


fzadrazil:
# Bug

**Describe the bug**
Trying to add this ArcGIS Service by its URL to the map doesn't work. No layers are loaded
https://ags.plzen.eu/arcgis/rest/services/GIS_Historicke/GIS_HIS_Cisarske_otisky/MapServer

**To Reproduce**
1. Go to the Add data panel
2. Select URL > ArcGIS Map Service
3. Insert the mentioned URL and connect to it
4. No layers are loaded and adding to the map is not possible

**Expected behavior**

**Screenshots**

**Additional context**

DailisLangovskis:
@fzadrazil I will look into it in the middle of this week at some point!

PR #3518, opened 2022-11-08
fix: Don't cache getCapabilities, if response is error or no data was…

DailisLangovskis:
## Description

Arcgis services did not load any layers from the url. It was most likely because of some error while waiting for the response from the server. In Such case we should not store in cache this error response. Also if no layers are returned from the request, we should also no cache it.

refs #3515 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist


raitisbe:
The service now returns correct response so can't really replicate, but this enhancement seems still useful. Thanks, gonna merge

raitisbe (approved):

PR #3529, opened 2022-11-09
Basemap compo

### LINKED ISSUE #3307: Use map composition as a source for basemap gallery
# Feature Request

I am migrating map atlas portal which contains tens of map projects = compositions. Each of those maps has the same set of basemaps (there is 10 basemaps), only the default basemap changes through the projects.

It would be nice and helpful to create one "basemaps map composition" which would contain the set of all the basemaps and then be able to use this composition as a source of the basemaps for hslayers basemap gallery. There would be a new configuration parameter
```
basemaps: {
  sourceUrl: '<link to the map composition file>',
  default: 'name of the default visible basemap'
}
```

Copy all the basemaps to all of the map compositions.

FilipLeitner:
## Description

Use map composition as a source for basemaps using 
````
appConfig.base_layers{
  url: string, 
  default: string, //Base layer which should be visible by default
}
````

- Reload defaultComposition layers on map reset

fixes #3307 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/map/map.service.ts
+++ b/projects/hslayers/src/components/map/map.service.ts
@@ -2,7 +2,6 @@
 import {Injectable, Renderer2, RendererFactory2} from '@angular/core';
 
 import ImageWrapper from 'ol/Image';
-import {Vector as VectorLayer} from 'ol/layer';
 import projx from 'proj4';
 import {
   Cluster,
@@ -35,6 +34,7 @@ import {Feature, ImageTile, Kinetic, Map, MapBrowserEvent, View} from 'ol';
 import {Geometry} from 'ol/geom';
 import {Group, Layer, Tile} from 'ol/layer';
 import {Projection, transform, transformExtent} from 'ol/proj';
+import {Vector as VectorLayer} from 'ol/layer';
 import {platformModifierKeyOnly as platformModifierKeyOnlyCondition} from 'ol/events/condition';
 import {register} from 'ol/proj/proj4';
 
@@ -822,6 +822,7 @@ export class HsMapService {
       }
       this.addLayersFromAppConfig(defaultLayers, visibilityOverrides, app);
     }
+    this.hsEventBusService.loadBaseLayersComposition.next({app});
   }
 
   /**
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Review by fzadrazil — CHANGES_REQUESTED** [review]
Few issues still present:
* when more basemap options expanded, it is hidden
image
* OSM is loaded even when other layer is set as default causing two overlapping basemaps rendered in the map

**Comment by fzadrazil** [comment]
One last issue. When you open the gallery and:
* click on the current basemap (nothing happens, that's fine)
* now click on another basemap (that one is now selected and the gallery closes)
* open the gallery again - no basemap is marked as selected

**Comment by FilipLeitner** [comment]
Oh very well. I modiefied changeBaseLayerVisibility function a bit so I guess I messed up something a bit. Thanks for spotting it. Will fix

File: projects/hslayers/src/components/map/map.service.ts [hunk 1]
```diff
@@ -1198,19 +1198,19 @@ export class HsMapService {
...
-    this.getLayersArray(app).forEach((lyr) => {
-      if (
-        getFromComposition(lyr) ||
-        getRemovable(lyr) === undefined ||
-        getRemovable(lyr) == true
-      ) {
-        to_be_removed.push(lyr);
-      }
-    });
+  removeCompositionLayers(force: boolean, app: string): void {
```
jmacura, line 1204: Could you consider changing the method's signature to
```suggestion
  removeCompositionLayers({force, app}: {force: boolean; app: string}): void {
```
also perhaps while setting some sane defaults, effectively making the params optional.

It is so much more readable to call it like `removeCompositionLayers({force: true})`
instead of `removeCompositionLayers(true, 'app')`

FilipLeitner: Good point, will refactor.

FilipLeitner: Take a look here 0bf8d75babe1ef01270508bc50d43ac93ffab1cc. 

Made force param optional effectively setting its default to false.
Also tried documenting it properly. More or less it works. Param description is infered from interface
image

jmacura: Wow, so much profi with the interface! Thanks!

jmacura:
Ready to merge, @FilipLeitner ?

Issue #3586, opened 2022-12-05
Show style visualization (legend) in rule list at styler panel.


raitisbe:
# Feature Request

**Is your feature request related to a problem? Please describe.**
If style rule names are not descriptive enough its hard to find the corresponding rule for a given visual representation.

**Describe the solution you'd like**
A small icon in front of each style rule to visualize that particular style

PR #3617, opened 2022-12-08
feat: Display legend for each rule in styler panel

raitisbe:
## Description

image

#3586 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

## Checklist


jmacura (commented):
I really like this feature!

Some thought:
Perhaps changing the background under the legend and several pixels around to white? With the blue background the mix of transparent colours is a bit misleading. In extreme case with `opacity: 0` it may look like blue.
image

jmacura (commented):
In some more complex cases, the border thickness is malformed (overflows from the span?). Not sure if there is an easy fix. Not a critical issue.
image

jmacura (commented):
What is more serious, the opacity is not taken into account when the styler is opened (the same goes for the legend panel). It is reflected only after changing the opacity value with the slider.
image
image

raitisbe:
> What is more serious, the opacity is not taken into account when the styler is opened (the same goes for the legend panel). It is reflected only after changing the opacity value with the slider. image image

Thats interesting, will have to look into it

raitisbe:
> In some more complex cases, the border thickness is malformed (overflows from the span?). Not sure if there is an easy fix. Not a critical issue. image

Yes, i think I will leave it be. That for geostyler to solve

raitisbe:
> Perhaps changing the background under the legend and several pixels around to white

White would be best, yes

raitisbe:
I made some fixes and tweaks. Could you check the opacity thing again @jmacura ? For me the slider seemed to be correct. Only the opacity in legend panel was not being applied, because of mismatch how `fillOpacity` and ordinary `opacity` is being handled.

jmacura:
> I made some fixes and tweaks. Could you check the opacity thing again @jmacura ? For me the slider seemed to be correct. Only the opacity in legend panel was not being applied, because of mismatch how `fillOpacity` and ordinary `opacity` is being handled.

Not any better.
image
I am unsure now if the issue is with the opacity or with the fact that there are two fill symbolizers used. The opacity of the second symbolizer is ignored until its value changes. Other aspects of the symbolizer are rendered correctly though (stroke color, stroke width, fill color, ...). Not sure if we can do something about it or if the geostyler is to blame again...

jmacura:
> > Perhaps changing the background under the legend and several pixels around to white
> 
> White would be best, yes

Good idea with the background transition. The thing is ... I do not like it 😆 And that's because we do not use color transitions anywhere else in the GUI. So I have played a little and came up with this.
_Original:_
image
_Mine:_
image

What do you think?
I am pushing the commit, you can take it or revert it as you like.

raitisbe:
Now it loks too much like a textbox imho. Up for a vote?

raitisbe:
Nevertheless I think its good to have a different style for rules apart from symbolizer headings.

raitisbe:
image
Made it a real text box and in the process got rid of the other name field when opening the rule.

raitisbe:
Thanks for assistance ;). I'm gonna merge this PR and the other sld related one to be able to rebase the package-lock. We can work on the look&feel afterwards.

jmacura:
> image Made it a real text box and in the process got rid of the other name field when opening the rule.

Wow, I really like this. Elegant solution! 💯

fzadrazil:
That final solution looks great! Superb new feature 👍

PR #3705, opened 2023-01-13
Fix composition/add-data item list overflow triggering map extent change

FilipLeitner:
## Description

Excess height of add-data and composition lists were triggering `extentChange` event (by changing map width when the scrollbar appeared)
Fixed by moving scroll deeper into DOM - now only panel content shrinks.

Compositions panel:
compositions

For add-data panel it was a bit different because of different DOM structure.  The issue is fixed, but scrolling behaviour of some of the elements is a bit different than expected I'd say (different from the composition panel at very least). 
I was not able to keep both add-data tabs (catalogue,URL,file) and catalogue header visible.. Its one or the other.

I chose the header because the other option affects not only catalogue but other tabs (for example when adding WMS layer, tabs would be visible when scrolling). Not that it would look bad its just change from what we have atm.
See examples below 
Tabs:
adddata-tab

Header:
adddata-header

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22243,6 +22243,7 @@
     },
     "node_modules/vega-embed/node_modules/yallist": {
       "version": "4.0.0",
+      "extraneous": true,
       "inBundle": true,
       "license": "ISC"
     },
@@ -40243,7 +40244,8 @@
         },
         "yallist": {
           "version": "4.0.0",
-          "bundled": true
+          "bundled": true,
+          "extraneous": true
         }
       }
     },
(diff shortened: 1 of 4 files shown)


raitisbe:
I think it looks very nice now. Thanks

raitisbe (approved):

PR #3736, opened 2023-01-24
fix: Load external wfs layers without add layers panel

raitisbe:
## Description
HsCompositionsParserService.jsonToLayer doesnt return a layer when lyr_def.protocol?.format == 'hs.format.externalWFS' is used. This prevents external WFS layers in whiteboard (they are handled like oridinary vector layers by mw library, not haslayers instead)
This is fixed in this PR by utilizing common/layers/hs.source.WfsSource

Had problems with SRS because corrdinate pairs where switched in BBOX and in GML response. Tried to fix it for EPSG:4326, but special case for 4258 still remains 

```
if (/*srs.includes('4326') || */ srs.includes('4258')) {
```
Need to check that

This also supports a new `version` attribute in the composition schema which specifies VERSION param used in WFS requests:
```
"protocol":{
                    "url":"https://hub4everybody.com/geoserver/jan_vrobel/wfs",
                    "format":"hs.format.externalWFS",
                    "version": "2.0.0"
                },
```

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/projects/hslayers/src/common/layers/hs.source.WfsSource.ts
+++ b/projects/hslayers/src/common/layers/hs.source.WfsSource.ts
@@ -48,14 +48,20 @@ export class WfsSource extends Vector<Geometry> {
         const srs = crs.toUpperCase();
 
         extent = transformExtent(extent, projection.getCode(), srs);
-        if (srs.includes('4326') || srs.includes('4258')) {
+        if (/*srs.includes('4326') || */ srs.includes('4258')) {
           extent = [extent[1], extent[0], extent[3], extent[2]];
         }
+        //https://gis.stackexchange.com/questions/30602/openlayers-wfs-flip-coordinates
+        //Do this, so feature coordinates would be in the right order. Without urn they aren't
+        //Can test with https://hub4everybody.com/geoserver/jan_vrobel/wfs
+        const srsWithUrn = srs.includes('urn')
+          ? srs
+          : 'urn:x-ogc:def:crs:' + srs;
         const params = {
           service: 'wfs',
           version: data_version, // == '2.0.0' ? '1.1.0' : data_version,
           request: 'GetFeature',
-          srsName: srs,
+          srsName: srsWithUrn,
           output_format: output_format,
           // count: layer.limitFeatureCount ? 1000 : '',
           BBOX: extent.join(',') + ',' + srs,
--- a/projects/hslayers/src/common/layers/hs.source.WfsSource.ts
+++ b/projects/hslayers/src/common/layers/hs.source.WfsSource.ts
@@ -44,24 +44,21 @@ export class WfsSource extends Vector<Geometry> {
         if (typeof output_format == 'undefined') {
           output_format = data_version == '1.0.0' ? 'GML2' : 'GML3';
         }
-
         const srs = crs.toUpperCase();
-
         extent = transformExtent(extent, projection.getCode(), srs);
-        if (/*srs.includes('4326') || */ srs.includes('4258')) {
+        if (
+          //https://gis.stackexchange.com/questions/30602/openlayers-wfs-flip-coordinates
+          //WFS version 1.x uses lat/lon order while version 2.x uses lon/lat order
+          data_version.startsWith('1') &&
+          (srs.includes('4326') || srs.includes('4258'))
+        ) {
           extent = [extent[1], extent[0], extent[3], extent[2]];
         }
-        //https://gis.stackexchange.com/questions/30602/openlayers-wfs-flip-coordinates
-        //Do this, so feature coordinates would be in the right order. Without urn they aren't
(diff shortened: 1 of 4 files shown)


FilipLeitner:
Changes in `createWFSLayer` function seem like a duplication. Adding external wfs from composition is indeed broken. We could just probably use https://github.com/hslayers/hslayers-ng/blob/e507c1fb38755998a49dc7cc796f4bd170ee07a5/projects/hslayers/src/components/add-data/url/add-data-ows.service.ts#L65
instead of proposed because in current state `createWFSLayer` was just incomplete extraction of that `connect` method.
To make it work with connect some fixes need to be done in url-wfs-service. We do not utilize checked and that's why its not working:
Getting empty collection here
https://github.com/hslayers/hslayers-ng/blob/e507c1fb38755998a49dc7cc796f4bd170ee07a5/projects/hslayers/src/components/add-data/url/wfs/wfs.service.ts#L116
because here
https://github.com/hslayers/hslayers-ng/blob/e507c1fb38755998a49dc7cc796f4bd170ee07a5/projects/hslayers/src/components/add-data/url/wfs/wfs.service.ts#L362

add_all is true (even tho its not supposed to be - adding just one layer) and layer.checked is never true if not using UI

raitisbe:
> Changes in `createWFSLayer` function seem like a duplication. Adding external wfs from a composition is indeed broken. If I tracked that correctly, then we could probably just use
> 
They are very similar, but in layer-parser service I don't rely on appRef anymore and everything comes from layer_def which is according to composition schema, not our own type (casing of Title, usage of .style, path instead of folder)

FilipLeitner:
O I havent noticed those details, right. Will do a quick test on this branch before approving

FilipLeitner:
With the service provided in comment it worked. I also tried [this one](http://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer) and it behaves quite strangely.

Loads with no problem by UI. 
From composition it loads, no features ...need to pan to switched coordinates and only after that the features are loaded.
So it seems like no matter what format of EPSG we provide it still requires the switch.

See

[srsName: EPSG:4326 + BBOX: 12.3929412,47.235098,16.4870588,51.2109664,urn:x-ogc:def:crs:EPSG:4326](https://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer?service=wfs&version=1.1.0&request=GetFeature&srsName=EPSG:4326&output_format=GML3&BBOX=12.3929412,47.235098,16.4870588,51.2109664,urn:x-ogc:def:crs:EPSG:4326&typeName=ukm:Pokryt%C3%AD_%C3%9AKM)

[Same BBOX but srsName: urn:x-ogc:def:crs:EPSG:4326](https://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer?service=wfs&version=1.1.0&request=GetFeature&srsName=urn:x-ogc:def:crs:EPSG:4326&output_format=GML3&BBOX=12.3929412,47.235098,16.4870588,51.2109664,urn:x-ogc:def:crs:EPSG:4326&typeName=ukm:Pokryt%C3%AD_%C3%9AKM)

[Switched coordinates + srsName: urn:x-ogc:def:crs:EPSG:4326](https://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer?service=wfs&version=1.1.0&request=GetFeature&srsName=urn:x-ogc:def:crs:EPSG:4326&output_format=GML3&BBOX=47.235098,12.3929412,51.2109664,16.4870588,urn:x-ogc:def:crs:EPSG:4326&typeName=ukm:Pokryt%C3%AD_%C3%9AKM)

[Swtiched coordinates + srsName:EPSG:4326](https://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer?service=wfs&version=1.1.0&request=GetFeature&srsName=EPSG:4326&output_format=GML3&BBOX=47.235098,12.3929412,51.2109664,16.4870588,EPSG:4326&typeName=ukm:Pokryt%C3%AD_%C3%9AKM)

jmacura:
@FilipLeitner Could you please test if the last commit makes the WFS load properly working in all known scenarios? 🙏

FilipLeitner (approved):

jmacura:
> Tested with [this](http://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer) and [this](https://hub4everybody.com/geoserver/jan_vrobel/wfs)
> 
>     * add-data url OK
> 
>     * save to composition and load OK
> 
>     * Load random WFS from layman OK
>       So I guess it should be all right

Finally! 😅 We will see if it won't break some other scenario. Thanks.

PR #3837, opened 2023-02-20
Raster timeseries

### LINKED ISSUE #3755: Upload of raster time series
# Feature Request

Layman supports publishing time series of rasters since version 1.18. Rasters should be uploaded in a zip package and a time regex must be provided so Layman knows where to look for the data/time in the file names, eg. 

`time_regex=[0-9]{8}` would look for the set of eight numbers, yyyyMMdd by default.
If the format is specified as part of the regex parameter, eg. `[0-9]{8},format=yyyyddMM`, it would be interpreted based on this format.

**Additional context**
[Add any other context or screenshots about the feature request here.
](https://github.com/LayerManager/layman/blob/master/doc/rest.md#post-workspace-layers:~:text=array%20of%20strings-,it%20is%20allowed%20to%20publish%20time%2Dseries%20layer,-by%20setting%20time_regex)

https://docs.geoserver.org/stable/en/user/community/cog/mosaic.html?highlight=app%20schema#timeregex-properties

FilipLeitner:
## Description
- Implement raster series upload to Layman

    Currently works with raster series files that:
      - matche one of the following regex datetime formats 
           -[0-9]{8}T[0-9]{9}Z eg 20220510T050948Z
           -[0-9]{8}T[0-9]{9} eg. 20220510T050948
       - consist of digits only or digits and separators  ., _, /, or -
       
    For the regex patterns listed  I dont really know wether these are the most used . I took those from [layman docs](https://github.com/LayerManager/layman/blob/v1.18.0/doc/rest.md#post-workspace-layers). 
     We can always add/edit this.

 - Allow all CRS from supported CRS list to be used with rasters.

closes #3755 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/test-app/src/hslayers-app/hslayers-app.component.ts
+++ b/projects/test-app/src/hslayers-app/hslayers-app.component.ts
@@ -392,7 +392,7 @@ export class HslayersAppComponent {
             },
           ],
           proxyPrefix: window.location.hostname.includes('localhost')
-            ? `${window.location.protocol}//${window.location.hostname}:8085/`
+            ? `${window.location.protocol}//127.0.0.1:8085/`
             : '/proxy/',
           panelsEnabled: app.panelsEnabled,
           mapSwipeOptions: {
--- a/cypress/tests/add-data.cy.ts
+++ b/cypress/tests/add-data.cy.ts
@@ -23,7 +23,7 @@ describe('Hslayers application', () => {
 
     //it('Wms capabilities should be retrieved', () => {
     cy.get(`hs-url-wms hs-common-url input`).type(
-      'https://hub.lesprojekt.cz/geoserver/leitnerfilip_wms/ows'
+      'https://watlas.lesprojekt.cz/geoserver/filip_wms/ows'
     );
     cy.get(`hs-url-wms hs-common-url input + button`).click();
     cy.wait(1000);
--- a/projects/hslayers/src/assets/locales/cs.json
+++ b/projects/hslayers/src/assets/locales/cs.json
@@ -40,7 +40,7 @@
       "highlightAndSelect": "Označte a vyberte časovou značku v názvu",
       "inferredRegex": "Odvozený regulární výraz",
       "selectedString": "Vybraný řetězec",
-      "verifyInferredRegex": "Před pokračováním ověřte odvozený regulární výraz"
+      "verifyInferredRegex": "Před pokračováním ověřte a potvrďte odvozený regulární výraz"
     },
     "requestServiceLayers": "Vyžádat vrstvy",
     "saveToCatalogue": "Uložit do katalogu",
--- a/projects/hslayers/src/assets/locales/cs.json
+++ b/projects/hslayers/src/assets/locales/cs.json
@@ -123,7 +123,6 @@
     "useTiles": "Použít dlaždice",
     "Vector": {
       "addingFiles": "Přidávání {{featuresCount}} prvku do:",
-      "addSld": "Přidat soubor SLD/QML",
       "chooseFiles": "Vyberte soubory...",
       "chooseLayer": "Vyberte prosím vrstvu ze seznamu ...",
       "dragAndDropFiles": "Přetáhněte soubory",
@@ -150,7 +149,10 @@
       "pickIdParamPt1": "Která proměnná obsahuje",
       "pickIdParamName": "unikátní identifikátor",
       "pickIdParamPt2": "dotazovaných prvků?"
-    }
+    },
+    "addsld": "Přidat soubor SLD",
+    "addqml": "Přidat soubor QML",
+    "addsldqml": "Přidat soubor SLD/QML"
   },
   "COMMON": {
     "abstract": "Abstrakt",
--- a/projects/hslayers/src/assets/locales/cs.json
+++ b/projects/hslayers/src/assets/locales/cs.json
@@ -36,7 +36,11 @@
     "FILE": {
       "onlyOneZipFileCan": "Lze nahrát pouze jeden soubor ZIP najednou",
       "someOfTheUploadedFiles": "Nahrávaný soubor přesahuje maximální povolenou velikost 20MB",
-      "zipFileCannotBeUploaded": "Soubor archivu nemůže být nahrát současně s běžnými soubory"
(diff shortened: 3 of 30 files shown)


jmacura:
Wow, seems amazing to me! 👏 
I think it makes sense to make it a separate type of file, the workflow is different from a single image upload.

projects/test-app/src/hslayers-app/hslayers-app.component.ts
```diff
@@ -392,7 +392,7 @@ export class HslayersAppComponent {
             },
           ],
           proxyPrefix: window.location.hostname.includes('localhost')
-            ? `${window.location.protocol}//${window.location.hostname}:8085/`
+            ? `${window.location.protocol}//127.0.0.1:8085/`
```
jmacura, line 395: commited intentionally, @FilipLeitner ?

jmacura (approved):
I like the use of reactive forms! 💯 

Perhaps you would like to add a sk translation for "Raster time series"?
And please check my comment about the test-app.

We can merge then.

projects/test-app/src/hslayers-app/hslayers-app.component.ts
FilipLeitner: Yes, kinda intentional. Not sure what changed but when running e2e cypress was not able to access proxy under the 'localhost' hostname. If I remember correctly both 127.0.0.1 and 0.0.0.0 was working ...just localhost not. Tried few cypress setting but no success. 
And its pretty much the same thing so I felt its safe to be changed in this way

jmacura: Ack. Verified. Happens to me as well. So let's go this way now. Thanks for explaining.

fzadrazil (changes requested):
The UI looks great! That is how I imagined it should be.
The only thing left is the list of supported CRS. I can still select only from epsg:3857 or epsg:4326, but other should be provided in the drop down. See the list of supported CRS [here](https://github.com/LayerManager/layman/blob/v1.18.0/doc/env-settings.md#LAYMAN_INPUT_SRS_LIST). The same applies to the single raster files.

jmacura:
> The UI looks great! That is how I imagined it should be. The only thing left is the list of supported CRS. I can still select only from epsg:3857 or epsg:4326, but other should be provided in the drop down. See the list of supported CRS [here](https://github.com/LayerManager/layman/blob/v1.18.0/doc/env-settings.md#LAYMAN_INPUT_SRS_LIST). The same applies to the single raster files.

Oh, that is strange. I can see all supported CRSs in the list... 
image

FilipLeitner:
> The UI looks great! That is how I imagined it should be. The only thing left is the list of supported CRS. I can still select only from epsg:3857 or epsg:4326, but other should be provided in the drop down. See the list of supported CRS [here](https://github.com/LayerManager/layman/blob/v1.18.0/doc/env-settings.md#LAYMAN_INPUT_SRS_LIST). The same applies to the single raster files.

Was most likely caused by failed layman's version check (no proxy running?). We used just part(3857,4326) of supported crs as a default value. It was updated only after the layman's version was verified to be above 1.16.
Since we support all CRSs from the list I turned the other way around.

fzadrazil:
Interesting, I am pretty sure the proxy was running, I would not be able to upload the rasters without it. But it seems to work today (with and without the last fix). So I guess it just needed to get some sleep  :-)

I have some more remarks. How does the timeregex validation work exactly? When I use the data as in the screenshot, the regex is suggested as [0-9]{15} which is obviously not right and will fail during the publication on Layman. But the component says it is valid.
image

As soon as the publication fails, Layman returns an error like this
```
"file": {
    "error": {
      "code": 48,
      "detail": {
        "expected": "All main data files match time_regex parameter",
        "message": "File does not match time_regex.",
        "unmatched_filenames": [
          "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220510T050948_20220510T051013_043144_05271A_E359_RVI4S1.tif",
          "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220522T050948_20220522T051013_043319_052C50_287D_RVI4S1.tif",
          "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220603T050949_20220603T051014_043494_053176_E5BF_RVI4S1.tif",
          "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220615T050950_20220615T051015_043669_0536B1_1277_RVI4S1.tif"
        ]
      },
      "message": "Wrong combination of parameters"
    },
```
But there is only "48" in the error toast message.
image

FilipLeitner:
Refarctored regex patterns it now accept
[0-9]{8}T[0-9]{9}Z
[0-9]{8}T[0-9]{9}
[0-9]{8}T[0-9]{6}Z
[0-9]{8}T[0-9]{6}
And digits separated by one of ., _, /, or -

- reworded regex verification step to 'confirmation' to make it clearer that it's the user who's supposed to make sure that generated regex fits is correct.

fzadrazil:
> As soon as the publication fails, Layman returns an error like this
> 
> ```
> "file": {
>     "error": {
>       "code": 48,
>       "detail": {
>         "expected": "All main data files match time_regex parameter",
>         "message": "File does not match time_regex.",
>         "unmatched_filenames": [
>           "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220510T050948_20220510T051013_043144_05271A_E359_RVI4S1.tif",
>           "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220522T050948_20220522T051013_043319_052C50_287D_RVI4S1.tif",
>           "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220603T050949_20220603T051014_043494_053176_E5BF_RVI4S1.tif",
>           "rvi4s1.zip/S1A_IW_GRDH_1SDV_20220615T050950_20220615T051015_043669_0536B1_1277_RVI4S1.tif"
>         ]
>       },
>       "message": "Wrong combination of parameters"
>     },
> ```
> 
> But there is only "48" in the error toast message. image

There still seems to be this one issue. But the workflow works great now even with styling. Good job!

fzadrazil (approved):

PR #3884, opened 2023-03-08
fix: Decode proxyfied url in case client send it that way

fzadrazil:
## Description

Some other clients using our proxy can send encoded URI (old HSLayers). Decode the URI in that case.

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

### DIFF
--- a/projects/hslayers-server/src/proxy.js
+++ b/projects/hslayers-server/src/proxy.js
@@ -30,6 +30,7 @@ require('http')
         res.end();
       }
       else {
+        req.url = decodeURIComponent(req.url);
         req.url = req.url.replace('/proxy', '');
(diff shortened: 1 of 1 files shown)


jmacura (approved):
I was afraid of some corner cases where names containing "decodable" string like %20 (e.g. "poll20%2023") got accidentally decoded while it shouldn't be. But by manual testing so far I can't simulate such a situation.

Hence the change is fine with me. Thanks!

FilipLeitner (approved):

jmacura:
This change was likely breaking after all.

I am receiving "400 Bad Request" from WFS service after this PR. Removing the decodeURIComponent line fixes it.
image

We have to rethink this @fzadrazil

PR #3898, opened 2023-03-10
hslayers-server: Properly encode query params in proxy

jmacura:
## Description

This is a hotfix of bug introduced in #3884

#3884

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers-server/src/proxy.js
+++ b/projects/hslayers-server/src/proxy.js
@@ -33,10 +33,9 @@ require('http')
         req.url = decodeURIComponent(req.url);
         req.url = req.url.replace('/proxy', '');
         const [base, queryParams] = req.url.split('?');
-        const params = querystring.decode(queryParams);
+        const params = querystring.parse(queryParams);
         req.url = base + '?' + querystring.encode(params);
         if (req.url.indexOf('api.geonames.org/searchJSON') > -1) {
-          const params = querystring.decode(req.url.split('?')[1]);
           if (
             typeof params.provider == 'undefined' ||
             params.provider == 'geonames'


jmacura:
Please, review this patch @fzadrazil or @FilipLeitner . Perhaps you will find another side-effects. Thanks!

projects/hslayers-server/src/proxy.js
```diff
@@ -32,6 +32,9 @@ require('http')
       else {
         req.url = decodeURIComponent(req.url);
```
FilipLeitner, line 33: Not sure what was the problem that lead you and Franta to adding decodeURIComponent line but isn't it redundant now?
Its decode - decode - encode for params.
So maybe if we want the path to be decoded it might be better to do it after the split.  It just feels a bit confusing/harder to understand this way to me

The only URL I had a problem with was [this ](http://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer?SERVICE=WFS&REQUEST=GetCapabilities) or [rather ](http://gis.kraj-jihocesky.gov.cz/gis/services/ukm/MapServer/WFSServer?service=wfs&version=1.1.0&request=GetFeature&resultType=hits&typeName=ukm:Pokryt%C3%AD_%C3%9AKM) (inferred feature count req for that service) and it works now

jmacura: > Its decode - decode - encode for params.

To be more precise, it is

1. decode whole URI
2. decode URL search params
3. encode URL search params

Step (2) is not there because of decoding, but because of parsing the URL search params into an object, which can be properly serialised (stringified) and encoded in step (3).

I agree this seems redundant. The whole process is there because the URL seach params shall be encoded but the rest of the URL ("base part") shall not be. Perhaps the code can be more simple, or we should even go without the `querystring` package, which is deprecated...

@fzadrazil may provide a test case for this. He mentioned that the problem was with using hslayers-server with the old (the not-NG?) hslayers.

fzadrazil (approved):

projects/hslayers-server/src/proxy.js
jmacura: I have switched from `querystring.decode()` to `querystring.parse()` which is an alias, but perhaps the code's intention is now a bit more clear, @FilipLeitner ?

jmacura: > 
> But I suppose once everything is migrated to NG, we can remove this decode/endode mess again.

I'd keep it anyway. You never know which client would like to use hs-proxy and in which form it will send the URI. I think this change is useful in general.

PR #4003, opened 2023-04-26
Reuse OWS connection functionality when loading composition layers (where possible)

FilipLeitner:
Starting with WFS as it brought this possiblity into attention.

Loading of external WFS was indeed broken before this fix https://github.com/hslayers/hslayers-ng/pull/3736
However I don't really see a reason for not using hsAddDataOwsService.connectToOWS as_
a) it basically duplicates add-data -> wfs -> getLayer sa mentioned [here](https://github.com/hslayers/hslayers-ng/pull/3736#issuecomment-1403313856) 
b) appRef is not a problem anymore https://github.com/hslayers/hslayers-ng/pull/3736#issuecomment-1404089735

I think that WMS/WFS/WMTS might be rewritten using add-data basically "just" by implementing option to set all the properties that are loaded from composition as a optional param for the layer loading pipeline


FilipLeitner:
Seems reasonable for WFS and WMTS.  
Not so clearly for WMS. 
- Would have to somehow deal with more layers in LAYERS param as connectToOWS adds layer one by one
-  Unnecessary getCapabilities call

Also an addition of possibility to set layer params for OWS might be handy

jmacura:
Could you rebase on current develop and resolve the conflicts @FilipLeitner ? I want to take a look at this PR again, hopefully merging it (finally).

jmacura (approved):

PR #4182, opened 2023-07-20
Calculate combined ArcGIS service layer extent only from valid extents

FilipLeitner:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/components/add-data/url/arcgis/arcgis.service.ts
+++ b/projects/hslayers/src/components/add-data/url/arcgis/arcgis.service.ts
@@ -40,7 +40,7 @@ export class HsUrlArcGisService implements HsUrlTypeServiceModel {
     public hsAddDataUrlService: HsAddDataUrlService,
     public hsAddDataCommonService: HsAddDataCommonService,
     public hsLayerUtilsService: HsLayerUtilsService,
-    public hsToastService: HsToastService
+    public hsToastService: HsToastService,
   ) {
     this.setDataToDefault();
   }
@@ -68,7 +68,7 @@ export class HsUrlArcGisService implements HsUrlTypeServiceModel {
    * @param wrapper - Capabilities response wrapper
    */
   async listLayerFromCapabilities(
-    wrapper: CapabilitiesResponseWrapper
+    wrapper: CapabilitiesResponseWrapper,
   ): Promise<Layer<Source>[]> {
     if (!wrapper.response && !wrapper.error) {
       return;
@@ -82,7 +82,7 @@ export class HsUrlArcGisService implements HsUrlTypeServiceModel {
       if (this.hsAddDataCommonService.layerToSelect) {
         this.hsAddDataCommonService.checkTheSelectedLayer(
           this.data.layers,
-          'arcgis'
+          'arcgis',
         );
         return this.getLayers();
       }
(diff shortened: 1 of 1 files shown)


projects/hslayers/src/components/add-data/url/arcgis/arcgis.service.ts
```diff
@@ -258,20 +258,27 @@ export class HsUrlArcGisService implements HsUrlTypeServiceModel {
...
       );
-      const layersExtents = layersCaps.map((l) => {
-        const extent = l.response.extent;
-        return this.transformLayerExtent(extent, this.data);
-      });
+      const layersExtents = layersCaps
+        .map((l) => {
+          if (
+            Object.values(l.response.extent)
+              .filter((v) => !(v instanceof Object))
```
jmacura, line 275:
```suggestion
              .filter((v) => !isNaN(v)
```

In a rare scenario, a number can be an instance of Object (Number initialized via the new Number())! It probably cannot happen in this case, but I feel using this comparison is generally unsafe. Additionally, it will interpret the `undefined` and `null` values also as "not instanceof Object" which is undesired. I would prefer isNaN comparison for these reasons. It will still fail for `Date` and `null`, but will solve the `undefined`. See this fiddle: https://jsfiddle.net/42ecxmkb/

jmacura (changes requested):

projects/hslayers/src/components/add-data/url/arcgis/arcgis.service.ts
FilipLeitner: Thanks for pointing it out. It really was not a good  https://github.com/hslayers/hslayers-ng/pull/4182/commits/2731938e8f48c4c8aeb46957440b34e45b9a79c4

FilipLeitner:
@fzadrazil Regarding the second issue.. Service caps include only combined extent so its the best to use it while display service as a whole. However, when displaying lets say 5 out of 10 layers we have no other option to get proper extent then querying  individual layers and calculating the extent as we did. 
We may try to to lower the limit for 'fullExtent' use from 100% to lets say 80-85 but it will be dependand on luck whether the extent needed to include possible outlayer or not.

FilipLeitner:
> @fzadrazil Regarding the second issue.. Service caps include only combined extent so its the best to use it while display service as a whole. However, when displaying lets say 5 out of 10 layers we have no other option to get proper extent then querying individual layers and calculating the extent as we did. We may try to to lower the limit for 'fullExtent' use from 100% to lets say 80-85 but it will be dependand on luck whether the extent needed to include possible outlayer or not.

https://github.com/hslayers/hslayers-ng/pull/4182/commits/95cc1e777df3fc8e07e60163f2da2624a7e10dcd

jmacura (approved):
Thanks for the change.
95cc1e777df3fc8e07e60163f2da2624a7e10dcd is a pure guessing, but whatever 🤪  Fine with me. Ready to merge?

fzadrazil:
Well, it is guessing :-) But I guess three requests is just the most we would like to make in that case.

PR #4246, opened 2023-08-14
Ignore WMS extent

### LINKED ISSUE #4201: Suppress extent limit for WMS layers
# Feature Request

To mitigate two issues:

1.  a layer with symbology, when WMS ignores the size of the symbols and does not take them into account when calculating the extent. Hence the symbols are cropped on the edges.
2. for layers added from Layman which use external URI to PostGIS database, the update of extent is not triggered after edit. Thus, when a user is adding features outside of current extent, this is not reflected and the features are invisible (because they are not requested by OpenLayers)

Add a checkbox to LayerManager saying "ignore layer extent" which will set the extent to undefined and force the layer to request GetMap for any part of the world. Disabled by default. Enabled by default for Layman/PostGIS layers.

Make the extent "a bit" larger for layers with symbology. This is a pure guessing however...
Instead of checkbox, implement slider saying "enlarge layer extent by N units"

**Additional context**
https://redmine.ccss.cz/issues/9578

FilipLeitner:
## Description

Add 'Igonre extent' toggle for WMS layers.

closes #4201 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/src/common/get-capabilities/wms-get-capabilities-response.interface.ts
+++ b/projects/hslayers/src/common/get-capabilities/wms-get-capabilities-response.interface.ts
@@ -114,7 +114,15 @@ export type WmsDimension = {
   current?: boolean;
   values: string | string[];
 };
-export type WmsLayer = {
+
+/***
+ * HsLayers extension of default WMS Layer object
+ */
+export type HsLayersWmsLayer = {
+  maxResolution: number;
+};
+
+export type WmsLayer = HsLayersWmsLayer & {
   Name?: NameType;
   Title: TitleType;
   Abstract?: AbstractType;
--- a/projects/hslayers/src/common/get-capabilities/wms-get-capabilities-response.interface.ts
+++ b/projects/hslayers/src/common/get-capabilities/wms-get-capabilities-response.interface.ts
@@ -115,14 +115,15 @@ export type WmsDimension = {
   values: string | string[];
 };
 
-/***
+/**
  * HsLayers extension of default WMS Layer object
  */
-export type HsLayersWmsLayer = {
+// eslint-disable-next-line no-use-before-define
+export type HsWmsLayer = WmsLayer & {
   maxResolution: number;
 };
 
-export type WmsLayer = HsLayersWmsLayer & {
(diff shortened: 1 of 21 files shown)


jmacura (changes requested):
Works great for layers added via URL. Does not work for layers added via HsConfig.

Also, I would change "extentStash" to "originalExtent" or "sourceExtent" or "wmsExtent" or "extentFromWms" or something alike..

projects/hslayers/src/common/layer-extensions.ts
```diff
@@ -56,6 +57,7 @@ const VIRTUAL_ATTRIBUTES = 'virtualAttributes';
 const WFS_URL = 'wfsUrl';
 const WORKSPACE = 'workspace';
 export const DOM_FEATURE_LINKS = 'domFeatureLinks';
+const WMS_EXTENT_STASH = 'wmsExtentStash';
```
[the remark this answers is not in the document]
FilipLeitner: Indeed

projects/hslayers/src/common/get-capabilities/wms-get-capabilities-response.interface.ts
```diff
@@ -114,7 +114,15 @@ export type WmsDimension = {
   current?: boolean;
   values: string | string[];
 };
-export type WmsLayer = {
+
+/***
+ * HsLayers extension of default WMS Layer object
+ */
+export type HsLayersWmsLayer = {
```
jmacura: I would defined the types the other way around...

jmacura (approved):
See last commit (rename of types).

Seems to work for init layers as well now 👍

projects/hslayers/src/common/get-capabilities/wms-get-capabilities-response.interface.ts
FilipLeitner: I was thinking about it as a hslayers extension to the base response, but yes this makes more sense. Thanks for the change

PR #4403, opened 2023-10-13
Hotfix addurl

FilipLeitner:
Missed that layerOptions are optional. Original change errors when not loading from composition.


jmacura (approved):

jmacura:
Thanks for the fix, I didn't realise that either.

PR #4414, opened 2023-10-17
Compositions refactoring & ArcGIS REST fixes (post workshop fixes)

FilipLeitner:
## Description

Fixes for some of the bug found during previous workshop..but mostly compositions rework.
Refactor to useangulars reactive forms + parsing of existing access_rights

~~Not tested with CSW  yet~~

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist


jmacura (commented):
👏 my hands for the many Doc comments :-)

I also fancy the use of reactive forms.
But I consider commits e4cc856275efa3037cd6c1df0493e5dbf61b1174 and 88377434f8fb14c5f736dd6b4e06810b68697909
to be a breaking changes. Albeit `hs-save-map-dialog-save` component was not used by hslayers-ng itself, it could have been used by a host application, hence it is potentially breaking. Similarly, the change of logic/removal of `compositionLoading` and `currentComposition` subjects can be a breaking change for hslayers-ng consumers. If you could separate these two changes/commits into a separate PR, I would be grateful and merge this right away. Otherwise I propose to wait 'till version 13.

jmacura:
So as discussed today, we will merge this and aim for the 13.0 release. The next Angular version is about to be released on November 6 so we may eventually make our release shortly after that.

projects/hslayers/src/components/core/event-bus.service.ts
```diff
@@ -131,10 +134,8 @@ export class HsEventBusService {
   /**
    * Fires when composition is downloaded from server and parsing begins
    * replaces 'compositions.composition_loading'
-   * @deprecated
    */
-  compositionLoading: Subject<any> = new Subject();
-  currentComposition: BehaviorSubject<any> = new BehaviorSubject(null);
+  compositionLoading: Subject<string> = new Subject();
```
jmacura, line 138: This does not hold. The only occasion where compositionLoading is broadcasted is from https://github.com/hslayers/hslayers-ng/blob/0ad079b45c3bcff0b67f5543d4a24d35c0f52f1e/projects/hslayers/src/components/save-map/save-map-manager.service.ts#L239 where its content ("compInfo") is not a string, but an object. The type-checker does not produce an error only because it is retyped to `any`.
Not sure what to do with this Subject. Perhaps delete it as well? We do not subscribe to it anywhere...

jmacura (changes requested):

projects/hslayers/src/components/core/event-bus.service.ts
FilipLeitner: Thanks! I was testing few possible solutions while trying to get current currentCompositionRecord to work and just missed this one.
Will get rid of this one.

jmacura: Thanks, merging.

jmacura (approved):

PR #4859, opened 2024-04-19
No main entry

FilipLeitner:
## Description

Lib structure changes:
 - Expose hslayers lib functionality via `secondary entry points only`. Main entry point is empty
 - Flatten folder structure by moving `assets`, `css `and `test files` outside of the `src `folder
 - `src` folder was replaced by `core` which includes former layout, map components and core service. 
 
Core basically replaced hslayers main entry point while allowing to simplify directory

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [x] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist


projects/hslayers/test/hslayers.component.spec.ts
```diff
@@ -6,7 +6,10 @@ import {CustomTranslationService} from 'hslayers-ng/services/language';
 import {HsConfig} from 'hslayers-ng/config';
 import {HsConfigMock} from './config.service.mock';
 import {HsLayerUtilsService} from 'hslayers-ng/services/utils';
-import {HsMapHostDirective, HslayersComponent} from 'hslayers-ng';
+import {
+  HsMapHostDirective,
+  HslayersComponent,
+} from 'hslayers-ng/core/public-api';
```
jmacura, line 12:
```suggestion
} from 'hslayers-ng/core';
```

projects/hslayers/components/sidebar/sidebar.component.ts
```diff
@@ -133,7 +131,7 @@ export class HsSidebarComponent implements OnInit, OnDestroy {
       !this.HsLayoutService.sidebarExpanded;
     this.HsLayoutService.updPanelSpaceWidth();
     setTimeout(() => {
-      this.HsCoreService.updateMapSize();
+      this.HsEventBusService.updateMapSize.next();
```
jmacura, line 134: I know it was not changed in this PR, but I propose to follow a common convention and rename `updateMapSize` to `mapSizeUpdates`. What do you think?

jmacura (approved):
Thanks! Overall looks good to me.
One idea crossing my mind is why we don't put config.service also into /core? Does it make sense to import it separately? Or vice versa, does it make sense to use HSLayers without config..?

FilipLeitner:
> Thanks! Overall looks good to me. One idea crossing my mind is why we don't put config.service also into /core? Does it make sense to import it separately? Or vice versa, does it make sense to use HSLayers without config..?

Config is being imported all around the lib and with it being a part of core we would easily end up having some sort of circular dependency which is not even allowed on entry point level eg.  core - map(service) - core.  Basically it would not be possible to import config in any of the entrypoints which are used in core (layout, utils, language, share etc)

jmacura:
> 
> Config is being imported all around the lib and with it being a part of core we would easily end up having some sort of circular dependency as is not even allowed on entry point level eg. core - map(service) - core. Basically it would not be possible to import config in any of the entrypoints which are used in core (layout, utils, language, share etc)

Get that, thanks for clarification!

Issue #4973, opened 2024-05-31
Upgrade to Angular 18


jmacura:
Currently waiting for several dependency packages to upgrade and provide a compatible release (custom-webpack, ng-bootstrap).

jmacura:
closed by #5003

PR #5003, opened 2024-06-14
Update Angular to version 18

jmacura:
## Description
Finally, all necessary packages are ready, so we can update to Angular 18.

#4973 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Dependency updates
- [x] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist


FilipLeitner:
Perhaps we could update  "lib": ["es2019", "dom"] in tsconfig.json. es2019 seem to be quite outdated. 

Features of es2022 seem to have a complete support (accros relevant browsers [see](https://caniuse.com/?feats=mdn-javascript_builtins_array_at,mdn-javascript_builtins_regexp_hasindices,mdn-javascript_builtins_object_hasown,mdn-javascript_builtins_error_cause,mdn-javascript_operators_await_top_level,mdn-javascript_classes_private_class_fields,mdn-javascript_classes_private_class_methods,mdn-javascript_classes_static_class_fields,mdn-javascript_classes_static_initialization_blocks)

From [es2023](https://caniuse.com/?feats=mdn-javascript_builtins_array_findlast,mdn-javascript_builtins_array_findlastindex,mdn-javascript_grammar_hashbang_comments,mdn-javascript_builtins_weakmap_symbol_as_keys,mdn-javascript_builtins_array_toreversed,mdn-javascript_builtins_array_tosorted,mdn-javascript_builtins_array_tospliced,mdn-javascript_builtins_array_with)  only Firefox seem to be last one to implement weakmap. We dont use it so it should not be an issue.

What do you think @jmacura  @fzadrazil ?

jmacura:
What would be the benefit for us?
We can always use all the latest features once they are supported in TypeScript language, no matter the output ES version. I always try to keep the output compiled JS code in the oldest version possible, in order to support the widest range of browsers.

FilipLeitner:
> We can always use all the latest features once they are supported in TypeScript language, no matter the output ES version

Well I am not entirely sure I understand. I believe that typescript already support all of the released ECMAScript features but without setting the newer target lib well be getting error such as
image

thats what I was aiming for . Also I think we should be safe to move from 5years old to at least 3 :D

jmacura:
> > We can always use all the latest features once they are supported in TypeScript language, no matter the output ES version
> 
> Well I am not entirely sure I understand. I believe that typescript already support all of the released ECMAScript features but without setting the newer target lib well be getting error such as
> image
> 
> thats what I was aiming for . Also I think we should be safe to move from 5years old to at least 3 :D
> 

Ah, my bad! I thought we are talking about the "target" option in tsconfig 😂 well, in that case, updating the "lib" option to something like "es2022" or "esnext" is quite all right I think.

jmacura:
> 
> Ah, my bad! I thought we are talking about the "target" option in tsconfig 😂 well, in that case, updating the "lib" option to something like "es2022" or "esnext" is quite all right I think.

It was illogical anyway, we had in tsconfig:
```
"target": "ES2022",
"lib": ["es2019", "dom"],
```

so we were targeting actually a newer version of JS than the version we were using in the codebase 😄

PR #5160, opened 2024-08-15
Extent buffer

FilipLeitner:
## Description

Fixes WMS extent buffering by using polynomial function to scale down the buffer factor. Overflowing extents are checked against EPSG:4087 bounds to prevent 'leaks'.

NOTE: invalid extents produce unexpected results

#5172 

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist


FilipLeitner:
Reliably buffering the extent seem to be quite a task.
- previous implementation was failing for local projections such as EPSG:5514 where simple comaprison with projected bounds of WGS84 is just not possible. apart from that the buffer size or large extents got above the bounds quite easily
- trying to transform enlarged extent into WGS84 doesnt really solves anything as the trasnformation returns NaN for overflowed latitudes and nonsense for overflowed longitude (because theres basically no scuh thing since we operate on globe).
- thats why I opted for manually defined bounds for each projection while trying to somehow manage the size of the buffer. This however creates a constraint and prevents us from displaying data outside the bounds, which is teoretically correct but from the technical point of view not so much (because theres no real issue with that...maybe if we were to automatically set maximum map extent along with the view it would be less confusing)

Stll work in progress tho ..needs some manuall testing or pereferably refactoring which would solve the problem without adding constraints or too much complexity 😄

FilipLeitner:
In [b26bcb8](https://github.com/hslayers/hslayers-ng/pull/5160/commits/b26bcb8c4ac4f792252f7d9ae5a510b8fdb04e90) Ive reverted many changes and implemented polynomial function to scale buffer factor based on extents' diagonal length instead of checking the bounds against the predefined extent.
Function used scales factor from 0.12  for small extents down to to 0 for the extents with diagonal length of more than 4000km (somewhat random number taht could be adjusted). 
It looks somewhat like this
image

FilipLeitner:
Well this is kinda insane.... our 4087 def is just wrong. No idea how. But it should not be 
```
    proj4.defs(
      'EPSG:4087',
      '+proj=cea +lon_0=0 +lat_ts=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
    );
```

but 

```
proj4.defs(
  'EPSG:4087',
  '+proj=eqc +lat_ts=0 +lat_0=0 +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs +type=crs',
);
```

honestly have no idea how I managed to screw this up :D

jmacura:
> Well this is kinda insane.... our 4087 def is just wrong. No idea how. It should not be
> 
> ```
>     proj4.defs(
>       'EPSG:4087',
>       '+proj=cea +lon_0=0 +lat_ts=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs',
>     );
> ```

Hmm... looks like [Lambert cylindrical equal-area](https://proj4.org/en/9.4/operations/projections/cea.html) projection to me 🤔 

> 
> honestly have no idea how I managed to screw this up :D

Shit happens 😄

FilipLeitner:
Wonder how would something like `map resolution` * `some pixel value` (ideally size of the symbol) work .

Because I mean...the resolution is a measure of how many map units per pixel are currently being displayed on the map...this would allow us to increas the extent in a terms of pixels. However, results would differ based on the current 'zoom level' of the map eg. zoom level at the time of layer creation.

fzadrazil:
> Wonder how would something like `map resolution` * `some pixel value` (ideally size of the symbol I guess) work .

Yep, but the big issue is still the unknown size of the symbol. If we know this, the problem would be solved. Let's try it as it is and see what occurs in real life :-)

fzadrazil (approved):

PR #5223, opened 2024-09-03
4229 -Wfs filter component

### LINKED ISSUE #4229: Create Filter WFS layer panel
# Feature Request

WFS layer can contain millions of features. The response from WFS server is thus slow and transferred data very large. By applying filter of attributes/attribute values, the amount of transferred and displayed data is reduced.

Create new panel for filtering WFS layers. Possibly also other filterable layers.

none

**Additional context**
Desired for EUHubs4Data project. See https://redmine.ccss.cz/issues/9569

### LINKED ISSUE #5294: Reorder comparison operators in WfsFilter and Styler panels
# Feature Request

Currently, the order of comparison operators presented to the user is: 'AND', 'OR', 'NOT', 'COMPARE'. In our left-to-right culture, this implies that 'AND' operator shall be used most often, while 'COMPARE' is the least important.

Reverse the order of comparison operators to: 'COMPARE', 'NOT', 'AND', 'OR'

Also, the simplest and most primitive operator 'COMPARE' could be pre-selected.

**Additional context**
image

FilipLeitner:
## Description

Create new panel for filtering WFS layers

- [x]  reuse the filtering functionality used in styler to construct the filters and then parse it via OL filter functions to the form usable in `[writeGetFeature](https://openlayers.org/en/latest/apidoc/module-ol_format_WFS-WFS.html#writeGetFeature)
- [x] adjust WFS source to use writeGetFeature along with filters created in WFS filter component
- [x] Fill information from `describeFeatureType` into component
- [x] Figure out WFS layer geometry column
- [x] Use it in filter
- [x] Make component a proper main panel
- [x] Styling

closes #4229
closes #5294 

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist


FilipLeitner:
Rough initial state just for the imagination
image

FilipLeitner:
Mostly ready. Missing
- [ ] Tests for wfs-filter component
- [ ] Shortcut pointing to WFS Filter component to layerEditor (tought we could input but its not that straigt forward with main panels so I skipped for the time being)
- [ ] More 'advanced' input options, range filters etc. how to handle it in UI....UX building some combinations are not the best. A good example would be a range of values where user has got to create both boudnaries individually or the fact that its not possible to type in value which could be usefil for += filters. 
- [ ] Following this, we could add an improvement for a styler filters in a form of parsed feature values to match wfs-filter. Currently only attribute names are prefilled in styler

jmacura:
One of the open questions I have remembered from the previous implementation was: How to store the filter information in the map composition?

UX scenario: User adds several WFS layers, sets filters and styles. Once the composition is saved and re-opened, they expect the map to look the same. Without storing the filter params along the layer/composition, the user might get a pretty different map.

FilipLeitner:
- [ ] Shortcut pointing to WFS Filter component from layerEditor (tought we could just use input but its not that straigtforward with main panels so I skipped for the time being)
- [ ] Store filters in composition

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
```diff
@@ -0,0 +1,325 @@
...
+   */
+  async selectLayer(layer: HsLayerDescriptor | null) {
+    this.selectedLayer.set(layer);
+    if (!layer) {
+      return;
+    }
+
+    const wfsUrl = getWfsUrl(layer.layer);
+    const url = this.hsUtilsService.proxify(
+      `${wfsUrl}?service=WFS&request=DescribeFeatureType&version=2.0.0&typeName=${layer.layer.get('layerName')}`,
```
jmacura, line 125: Why do you operate with `layerName` property? Existing property `name` is not sufficient?

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
```diff
@@ -0,0 +1,325 @@
...
+   * Selects a layer and updates the filter service
+   * @param layer The layer to select
+   */
+  async selectLayer(layer: HsLayerDescriptor | null) {
+    this.selectedLayer.set(layer);
+    if (!layer) {
+      return;
+    }
+
+    const wfsUrl = getWfsUrl(layer.layer);
```
jmacura, line 123: To ensure the WFS URL will be valid, I would trim any part after `?` mark, if it already exists in the `wfsUrl` property. For the sake of the following line.

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
```diff
@@ -0,0 +1,325 @@
...
+    HsFiltersComponent,
+    FormsModule,
+    AsyncPipe,
+    HsPanelHeaderComponent,
+    TranslateCustomPipe,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class HsWfsFilterComponent extends HsPanelBaseComponent {
+  name = 'wfsFilter';
```
jmacura, line 56: consider:

* Rename this new panel to just "filter", because it already exists as a value in `HsConfig.panelsEnabled` (kind of a similar panel was already implemented years ago and this param is a relict).
* Or remove the value "filter" from panelsEnabled for good.
* Or work with both values as with synonyms.

I have a weak preference for (1), but the two other options are also fine.

projects/hslayers/services/panel-constructor/panel-constructor.service.ts
```diff
@@ -90,6 +90,12 @@ export class HsPanelConstructorService {
       case 'tripPlanner':
         i = await import('hslayers-ng/components/trip-planner');
         break;
+      /**
+       * TODO: Check whether this get bundled correctly
```
jmacura, line 94: Still relevant?

projects/hslayers/common/filters/comparison-filter/comparison-filter.component.ts
```diff
@@ -0,0 +1,251 @@
...
+    FilterRangeInputComponent,
+  );
+  expanded = computed(() => this.filterRangeInput()?.expanded() ?? false);
+
+  private readonly OPERATORS = {
+    default: [
+      {value: '==', alias: '='},
+      {value: '!=', alias: '≠'},
+    ],
+    stringBased: [{value: '*=', alias: 'contains'}],
```
jmacura, line 77:
```suggestion
    stringBased: [{value: '*=', alias: '≈'}],
```

jmacura (commented):
Nice work, for sure!

I have a question, which you have already [mentioned in the previous comment](https://github.com/hslayers/hslayers-ng/issues/4229#issuecomment-2314826668):
Why parsing XML response instead of accepting JSON?

Then I have some observations when testing the panel:

- when loading GetPropertyValue, the loader disappears sooner, than the results are actually received by the browser. I can't see any timeout limit, so is there some default?
- when GetPropertyValues returns many values (4.82 MB of data in my case), the browser freezes.

Could you also please document new params at [HsConfig](https://github.com/hslayers/hslayers-ng/wiki/App-config-parameters) level, if any, and at [Layer/Source](https://github.com/hslayers/hslayers-ng/wiki/Layer-parameters) level?

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
FilipLeitner: I am not exactly sure, will take a look at this once more. Might be just some leftover from me trying experimenting with something.

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
FilipLeitner: Good idea,  sure.  Will do.

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
FilipLeitner: I'd like to keep 'wfs' in the name to communicate the  functionality properly. 
Regarding the `panelsEnabled.filter` I think it should be removed - panel no longer exists so this is no brainer for me.

So basically (2) :D

projects/hslayers/common/filters/comparison-filter/comparison-filter.component.ts
FilipLeitner: Yea this is much better 😆 thanks

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
FilipLeitner: There is bit of a mess with the layer name here
https://github.com/hslayers/hslayers-ng/blob/d9a306471c180f19130536cf1542e982889007da/projects/hslayers/services/add-data/url/wfs.service.ts#L514-L522

where we set both name and layerName (via options object) to the same value. I for some reason ended up using layerName. 
Theres also a note for thes property in type definition
https://github.com/hslayers/hslayers-ng/blob/d9a306471c180f19130536cf1542e982889007da/projects/hslayers/types/compositions/composition-layer-options.type.ts#L4
Anyways, name obtained via getName() should be perfectly fine and I guess preffered way to do so Ill make a change. Also it will make more sense to not double name related properties when defining wfs layer manually (in config for example)

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
FilipLeitner: [6e16401](https://github.com/hslayers/hslayers-ng/pull/5223/commits/6e16401c6c4514c1cdb42c266479501a321b1dbb)

projects/hslayers/components/wfs-filter/wfs-filter.component.ts
FilipLeitner: [80ccabf](https://github.com/hslayers/hslayers-ng/pull/5223/commits/80ccabf4be4efeb0f8fc23c75b5f9ca8353110d6)

FilipLeitner:
> 
> I have a question, which you have already [mentioned in the previous comment](https://github.com/hslayers/hslayers-ng/issues/4229#issuecomment-2314826668): Why parsing XML response instead of accepting JSON?
> 
Well JSON is rather sporadic Id say. Even we on layman dont support that see [here ](https://hub4everybody.com/hsl-layman/geoserver/franta/wfs?request=GetCapabilities&service=WFS&version=2.0.0)

 
> * when loading GetPropertyValue, the loader disappears sooner, than the results are actually received by the browser. I can't see any timeout limit, so is there some default?
> * when GetPropertyValues returns many values (4.82 MB of data in my case), the browser freezes.
>
I was not able to replicate any of these. However, I have added loading gif to prevent user from trying to pick attribute to be filtered before its ready

> Could you also please document new params at [HsConfig](https://github.com/hslayers/hslayers-ng/wiki/App-config-parameters) level, if any, and at [Layer/Source](https://github.com/hslayers/hslayers-ng/wiki/Layer-parameters) level?
Sure, thanks for pointing this out

jmacura (changes requested):
One more nitpick. If a user select a filter type "AND" and then deletes one of the filter lines, the UI allows them to click "Apply", but it ends with a silent error (yet a nicely descriptive one 👍 ).
image
image

Can we either:

- prevent removing the filters, so there are always at least 2 or
- prevent sending the "apply" call, if there are less then 2 filters? With some info message.

The same applies for the "OR" filter.

jmacura (approved):
image
Very nice!

jmacura:
Also fixes #5295

PR #5338, opened 2024-10-15
Fixes v15 - swipe layer render, map layers order, catalogue remove

FilipLeitner:
## Description

Fixes for few bugs:
-  messy rendering of map-swipe layers caused by swipe position not set correctly
-  Sort layers inside folders. Currently only folders were sorted, layers themselves where displayed in inital order
-  Layers not being removed from map when executed from catalgoue as a result of incorrectly set layer name. 

Along that :
- give priority to the property values in layerOptions over those in service defults
- Unify use_tiles and useTiles properties used in add-data

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist

### DIFF
--- a/projects/hslayers/components/map-swipe/map-swipe.service.ts
+++ b/projects/hslayers/components/map-swipe/map-swipe.service.ts
@@ -1,5 +1,6 @@
 import {Injectable, NgZone} from '@angular/core';
-import {buffer, filter, first, map} from 'rxjs';
+import {Observable, merge} from 'rxjs';
+import {buffer, filter, first, map, switchMap, take} from 'rxjs/operators';
 
 import {Layer} from 'ol/layer';
 import {Source} from 'ol/source';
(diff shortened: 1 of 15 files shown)


projects/hslayers/components/map-swipe/map-swipe.service.ts
```diff
@@ -100,6 +111,25 @@ export class HsMapSwipeService {
...
+  /**
+   * Get layer manager update along with the current mainPanel
+   */
+  private getLayerUpdateWithCurrentPanel(): Observable<{
+    layer: Layer<Source>;
+    panel: string;
+  }> {
+    return this.hsEventBusService.layerManagerUpdates.pipe(
+      //filter out nulls
+      filter((layer): layer is Layer<Source> => !!layer),
```
jmacura, line 123: What, are we in Python now? What is the `is` magic? 😮

projects/hslayers/services/layer-manager/layer-manager-folder.service.ts
```diff
@@ -199,21 +211,54 @@ export class HsLayerManagerFolderService {
...
+    return this.sortFoldersByZ(newState, false);
   }
 
   /**
    * Sorts folders by z-index.
    */
   private sortFoldersByZ(
     state: Map<string, HsLayermanagerFolder>,
+    debounce = true,
   ): Map<string, HsLayermanagerFolder> {
```
jmacura, line 223: I would prefer if the `debounce` param will be and optional property inside of an (optional) options object. The ultimate goal is more readable function calls, instead of `sortFoldersByZ(state, false)`, it would be `sortFoldersByZ(state, {debounce: false})`. What do you think? We have discussed this some time ago I think.

jmacura (approved):
LGTM, thanks. You can address the comment about function signature, if you agree.

projects/hslayers/services/layer-manager/layer-manager-folder.service.ts
FilipLeitner: Sure, I just keep forgetting. Sorry 😄

Issue #5622, opened 2025-03-04
Signals for auth state


FilipLeitner:
Current implementation of auth state and its reactivity is outdated unreliable and overly complicated. Refactor to signals/resource/rxjs

PR #5631, opened 2025-03-06
refactor: Auth state handling

FilipLeitner:
## Description

- simplified the auth state handling. Auth state is kept in common/layman.service (which could easily be renamed to auth.service) .
- it is separated from layman endpoint as `authState` while endpoint is kept as `layman$` or `layman` for signal
- easily accessible via computed properties isAuthenticated and user
- change is triggered by login and logout subjects or automatically at the app init
- in case auth session expires or token is revoked any request to layman endpoint that requires credentials will log the user out via authInterceptor
- updated current-user component 
image

- **simplified layman related code by removing the endpoint arguments from every method as there always has been only one layman endpoint which is also basically provider of authentication via `/current-user`**
- added test coverage for the service

#5622 

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [x] Yes
- [ ] No

## Checklist


FilipLeitner:
I tried to test layman related functionality such as:
- adding/removing layer (shp, geojson) - as WMS/WFS , with or without style
- geometry updates
- style updates
- composition load,/saving/renaming/updating

But as this change has changed a lot of code another series of user tests would be appreciated

projects/hslayers/services/add-data/vector/vector.service.ts
```diff
@@ -248,7 +248,8 @@ export class HsAddDataVectorService {
    */
   async addNewLayer(data: VectorDataObject) {
     if (!this.hsAddDataCommonFileService.endpoint) {
-      this.hsAddDataCommonFileService.pickEndpoint();
+      console.error('No endpoint available');
```
jmacura, line 251: Wouldn't it be better to show a toast message instead of a console error?

projects/hslayers/types/authentication.ts
```diff
@@ -0,0 +1,8 @@
+export interface AuthState {
```
jmacura, line 1: see above

```suggestion
export interface HsAuthState {
```

projects/hslayers/types/authentication.ts
```diff
@@ -0,0 +1,8 @@
...
+  user?: string;
+  authenticated?: boolean;
+}
+
+export interface AuthAction {
```
jmacura, line 6: see above

```suggestion
export interface HsAuthAction {
```

jmacura (changes requested):
Thanks for touching this complex code. A change is very welcomed.
I appreciate the test coverage too.

---

So far I have observed this when testing...

original window tab stays freeze in "signing in" state.
image

The newly opened window tab seems to work fine.

---

image
Untranslated string..?

projects/hslayers/services/add-data/common-file.service.ts
```diff
@@ -408,7 +379,7 @@ export class HsAddDataCommonFileService extends HsAddDataCommonFileServiceParams
     try {
       this.loadingToLayman = true;
       if (!this.endpoint) {
-        this.pickEndpoint();
+        throw new Error('No endpoint available');
```
jmacura, line 382: Oddly enough, I have hit this error today, even though I am logged in to a Layman endpoint successfully. So the `pickEndpoint()` probably had some meaning here.

jmacura (changes requested):

projects/hslayers/services/add-data/common-file.service.ts
jmacura: b5d03f4 shall address that -- the global `endpoint` property was always `undefined`. Moving it into constructor was also not optimal, due to the timing of async code. So I have moved it into a local scope.

FilipLeitner:
### NOTE: Complete  #5639 before merging this one

jmacura (approved):

PR #5803, opened 2025-06-12
Arcgis feature server support

FilipLeitner:
## Description

Please give a short description of the changes you propose. Please use prose and try not to be too technical, if possible. Make sure to mention people that you think should know about the PR.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [ ] No

## Checklist


jmacura (approved):
Very nice! Thanks for implementing this so quickly.
I propose one UI change, which is not directly related, but it hit me in the eyes...
The "Add as" label is so distant from the options, it is not immediately clear that it's just a label for the options on the right side...
image
For some time I thought it's a disabled button... So, instead, I propose to either move the options to the left like so:
image
Or to change the options from two buttons to a dropdown like the "Query format" and "SRS".
What is your opinion?

FilipLeitner:
Thats because for WMS there is another set of toggles. We could maybe disable the other options when applicable instead?
We would have to use two dropdowns otherwise
image

jmacura:
Hmm, right, right. Or we can emphasize the line somehow so it is more obvious that the label and options are related?
Like this, just nicer..?
image

PR #6053, opened 2025-10-31
Scale widget

FilipLeitner:
## Description

- update layer editor widget styles
- make scale widget togleable
- parse/store resolution for all type of layers into composition

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/projects/hslayers/components/layer-manager/widgets/scale-widget.component.html
+++ b/projects/hslayers/components/layer-manager/widgets/scale-widget.component.html
@@ -19,7 +19,7 @@
         <div class="input-group input-group-sm">
           <span class="input-group-text">1:</span>
           <input type="number" [id]="'min-scale-' + name" class="form-control" [(ngModel)]="minResolution"
-            name="minResolution" [placeholder]="placeholder" min="1" step="1" />
+            name="minResolution" [placeholder]="placeholder" min="0" step="1" />
         </div>
       </div>
 
--- a/projects/hslayers/components/layer-manager/widgets/scale-widget.component.html
+++ b/projects/hslayers/components/layer-manager/widgets/scale-widget.component.html
@@ -1,36 +1,42 @@
-<div class="form-group" [hidden]="!isScaleVisible()"><!-- TODO: Remove function call from template -->
-  <label>{{ "COMMON.scale" | translate }}</label>
-  @if (minResolutionValid()) {
-    <span><span class="ms-1">{{ "COMMON.from" | translate }}</span> 1:{{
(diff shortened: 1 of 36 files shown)


jmacura (commented):
The improved UI looks good, thanks!

How can I set the upper bound to infinity? For me, it is interpreted as 1:0 and thus nothing is shown. Do I have to set a random giant number?

Image

FilipLeitner:
> The improved UI looks good, thanks!
> 
> How can I set the upper bound to infinity? For me, it is interpreted as 1:0 and thus nothing is shown. Do I have to set a random giant number?
> 
> Image

Indeed this was not really intuitive. Made change so taht if user manually deletes the value it will set the limit to 0 for min and Infinity for max resolution.
Give it a try

jmacura (commented):
When minResolution is 0, the "hs-scale-summary" block won't show up. But the other way around it does.

Image

Image

Am I doing it wrong? "From" is minimum and "to" is maximum, right?

Also, I would label the whole component block "Scale restrictions" or "Scale limits" or "Scale visibility". Having just a "Scale" is not very clear. (QGIS has "Scale Dependent Visibility", which is a bit too long I think).

FilipLeitner:
image

projects/hslayers/components/layer-manager/widgets/scale-widget.component.html
```diff
@@ -1,36 +1,48 @@
...
+        </label>
+        <div class="input-group input-group-sm">
+          <span class="input-group-text">1:</span>
+          <input type="number" [id]="'max-scale-' + name" class="form-control" [(ngModel)]="maxResolution"
+            name="maxResolution" [placeholder]="placeholder" min="1" step="1" />
         </div>
       </div>
+    </div>
+
+    @if (minResolutionValid() && maxResolutionValid()) {
```
jmacura:
```suggestion
    @if (minResolutionValid() || maxResolutionValid()) {
```
Why not to show the summary in case any of the limits is valid?
That would be basically the same condition as `isScaleVisible()`. Perhaps we can use it directly...?
```suggestion
    @if (isScaleVisible()) {
```

jmacura (commented):
Thanks for the many improvements! I still struggle with the logic behind the hs-scale-summary though :-/

jmacura (approved):
Nice improvement!

PR #6246, opened 2026-04-23
Server rewrite with Hono and less dependencies

fzadrazil:
## Description

Complete rewrite of the hslayers-server components using Hono instead of Express.js and without the need of SQLite.

Two main motivations for the rewrite:
- SSRF vulnerabilities of the proxy
- tricky installation of SQLite packages

### DIFF
--- a/projects/hslayers-server/.gitignore
+++ b/projects/hslayers-server/.gitignore
@@ -4,3 +4,4 @@ data/
 .env.*
 !.env.example
 hslayers-server.db-journal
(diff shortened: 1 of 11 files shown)


jmacura:
Thanks for this massive rewrite. At a first glance, I can see there some legacy functions left in the code which are no longer in use. I hope to have some time to test the functionality next week. In general, we should test how it will work in the Liberec Region geoportal, where the setup is always tricky.

